PR #15, opened 2018-03-28
RFC: system node

dherman:
Summary:

> This is a high-level proposal, establishing how Notion should position itself relative to system installations of Node. Briefly: **Notion should recommend against being installed alongside system installations of Node.** On Unix operating systems, it's possible for Notion to offer a (non-default, opt-in) mode to forcibly override the system Node (with some caveats), but this should not be the recommended setup. Finally, this proposal also suggests future proposal ideas for helping users detect likely issues with an existing Notion installation, including other Node installations that may indicate conflicts.

[Rendered](https://github.com/notion-cli/rfcs/blob/e046568b0fb041b6c6255e50fd54351a98dddc84/text/0000-system-node.md)

### DIFF
--- a/text/0000-system-node.md
+++ b/text/0000-system-node.md
@@ -1,0 +1,108 @@
+- Feature Name: system_node
+- Start Date: 2018-03-26
+- RFC PR:
+- Notion Issue:
+
+# Summary
+[summary]: #summary
+
+This is a high-level proposal, establishing how Notion should position itself relative to system installations of Node. Briefly: **Notion should recommend against being installed alongside system installations of Node.** On Unix operating systems, it's possible for Notion to offer a (non-default, opt-in) mode to forcibly override the system Node (with some caveats), but this should not be the recommended setup. Finally, this proposal also suggests future proposal ideas for helping users detect likely issues with an existing Notion installation, including other Node installations that may indicate conflicts.
+
+# Motivation
+[motivation]: #motivation
+
+## System installs of Node are problematic
+
+System installs of Node in global locations like `/usr/bin` or `/usr/local/bin` are problematic. They have historically had [permissions issues](https://givan.se/do-not-sudo-npm/), and the advice on how to properly configure a system Node installation is [confusing at best](https://github.com/npm/npm/issues/3139#issuecomment-13358582).
+
+Global installs of Node also pose a security risk. Paths like `/usr/bin` are typically high precedence in a user's `PATH` variable, allowing for the possibility of accidental or even malicious shadowing of standard tools by third-party npm package executables. (As a relatively benign but still frustrating example, I have personally experienced a multi-hour debugging session where we eventually discovered a colleague's `which` had been overridden by the `node-which` package.)
+
+Maintaining additional Node installations alongside a system Node is especially problematic. For one, it's impossible for the second installation to override the system Node without placing third-party binaries _extremely_ early in the `PATH`, which carries the same security risks mentioned above. For two, even in this scenario, the user experience of global npm package binaries gets muddled: a global binary installed with the system Node may still get picked up if it hasn't been installed as a global in the current Node.
+
+For all the above reasons, **the best approach to using a Node version manager is to avoid having any other Node installations on the system**. Rather than attempting to hide the risks associated with co-installing along with a system Node, we should clearly recommend to users of Notion that if they have an existing system installation of Node, they should uninstall it first.
+
+## Unix and Windows
+
+In Unix OSes, we may want to consider still allowing the ability to opt into overriding a system installation. But since this incurs greater security risk and is generally less reliable, it should not be the recommended path. The installation steps should ask for confirmation and should recommend aborting and choose this as the default action.
+
+In Windows, the system `Path` variable takes precedence over the user `Path` variable---not dissimilar from how Unix environments place `/usr/bin` towards the beginning of `PATH`. But unlike Unix, Windows provides no general mechanism to circumvent this (whereas Unix users can simply prepend new directories to the front of `PATH`). This suggests that the Windows design has a stronger expectation that system installations take precedence over user installations.
+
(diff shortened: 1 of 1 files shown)


text/0000-system-node.md
```diff
@@ -0,0 +1,108 @@
...
+The idea that Notion installs the toolchain in the user's own space should not be too difficult to learn, since it's already the dominant model for many toolchain managers.
+
+# Details
+[details]: #details
+
+## Installation
+
+At installation time, the installer should check for:
+
+- Existing `node`, `npm`, `npx`, or `yarn` commands in the path.
```
mikrostew: These commands may also be in the path if the user has another node version manager installed (nvm, n, nodebrew, avn, &c.). Maybe a topic for a separate RFC, but should notion recommend against being installed alongside other version managers? Or is there a way that notion can coexist with other version managers?

text/0000-system-node.md
```diff
@@ -0,0 +1,108 @@
...
+Rather than attempting to circumvent standard practice or provide partial solutions that don't work for all expected contexts, the Windows Notion installer should simply fail if a system Node is detected.
+
+## Detecting system changes
+
+Detecting a system Node at installation time is insufficient since users or system administrators could install another Node installation on the system at any time. So we also need to provide users with the tools to detect these kinds of issues at any time after Notion has already been installed.
+
+# Pedagogy
+[pedagogy]: #pedagogy
+
+The simplest story to explain to users is: **don't install Notion alongside other Node installations**.
```
mikrostew: There should be a way for someone to try out Notion even if they are currently using a different version manager, without installing Notion globally (or at the beginning of the path). Possibly some way we can detect that there is another version manager installed, and give some messaging about how to disable the other one, or even automatically disable the other one.

text/0000-system-node.md
dherman: Good catch! This RFC should probably discuss that as well.

My sense is that it's going to be just as problematic to have another version manager active in your `PATH`. I'll chew on this while I commute to try to come up with some example problem cases…

text/0000-system-node.md
```diff
@@ -0,0 +1,108 @@
...
+
+In Windows, the system `Path` variable takes precedence over the user `Path` variable---not dissimilar from how Unix environments place `/usr/bin` towards the beginning of `PATH`. But unlike Unix, Windows provides no general mechanism to circumvent this (whereas Unix users can simply prepend new directories to the front of `PATH`). This suggests that the Windows design has a stronger expectation that system installations take precedence over user installations.
+
+Also unlike Unix, Windows developers do not expect environment variables to be configured by terminal bootup scripts. Windows users should be able to expect use cases like executing Node from "Start | Run" or in IDEs to work, which would not work with terminal configuration scripts.
+
+Rather than attempting to circumvent standard practice or provide partial solutions that don't work for all expected contexts, the Windows Notion installer should simply fail if a system Node is detected.
```
caridy: I have two main concerns:

* this imposes a high-tax on users to try Notion for the first time, because it will require them to rearrange everything that they already have set up. I think @mikrostew also reference to this problem, and maybe a local installation of some sort might workout better that just simply fail (at least for Windows users).
* using Notion as part of a much bigger toolchain (where the user is not installing Notion directly, but indirectly). This will force users to uninstall their global node if they ever want to install the "heroku toolbelt" or the "salesforce-cli", etc. If those toolchain want to rely on Notion for all their node stuff.

dherman: @caridy I don't disagree that it's an adoption hurdle. The issue is that I'm not sure I know how to solve it for Windows, given the fact that system `Path` overrides user `Path`.

We could play tricks with "junctions" (aka directory symlinks) but this doesn't work for parallel sessions (e.g. running in multiple terminals, or multiple users working on a shared Windows system).

I'm open to ideas!

dherman: I chatted with a friend at Microsoft. I'm going to investigate what nvs does, but I think most version managers just don't worry about the "leaked globals" problem. If that's true, then I think we can still do better without letting the perfect be the enemy of the good. Something like:

- At installation time, detect if there's a system Node in the system Path, and if so, install system "shim-shims" for whatever tools we find in the system Path (node, npm, any existing installed globals), which do nothing more than defer to the shims installed in the current user's Notion installation (if the current user is using Notion), and otherwise defer to the system node.
- Whenever the user runs `notion doctor`, update this set of shim-shims.

This isn't perfect in the sense that it can get out of sync with the system Node, in particular if people are installing and uninstalling globals a lot. But I think this is strictly better than what existing version managers do, and one plausible hypothesis of why they get away with this is that perhaps people often don't install global binaries in their system node.

I'll report back after digging more into what existing managers do.

benblank: Dave, and I discussed this a bit after a meeting today and we brainstormed a third possible approach, analogous to "git bash" on Windows.

During install, Notion could detect the existence of a global copy of Node and recommend removing it, but also provide the option of adding shortcuts which will open a Windows shell or PowerShell with the environment changed to remove the global Node from the PATH.  (If the user wants a similar experience in their Cygwin/MSYS/WSL shell, it should(?) be easy enough for them to modify their own appropriate dotfile.)  Notion could then run as normal, without even needing to be aware of the existence of an unmanaged copy of Node on the system.

This would allow "trial" adoption without absolutely requiring removal of global Node.  It would also avoid the issue of drifting out of sync with the global install, as any binaries from the global install would simply be unavailable (which may or may not be desirable, but does at least align with the message that Notion and a global Node install aren't meant to play nicely together).

benblank: I've just done some digging into how nvs handles global installs.  It works by manipulating the PATH of the running shell, so nvs-managed copies of Node aren't available in a particular shell until after nvs has been invoked:

```
C:\>where node && node --version
C:\Program Files\nodejs\node.exe
v8.11.1

C:\>nvs use 9
PATH -= C:\Program Files\nodejs
PATH += %LOCALAPPDATA%\nvs\node\9.11.0\x86

C:\>where node && node --version
C:\Users\535\AppData\Local\nvs\node\9.11.0\x86\node.exe
v9.11.0
```

dherman:
This RFC is basically all wrong at this point. :P I'm going to close it for now and maybe rewrite it from scratch.

PR #20, opened 2018-06-18
Exit codes

benblank:
> Exit codes are an important part of Notion's communication with external tools and a first-class part of the design of its user-friendly error messages. We need a strategy for ensuring they are used and communicated consistently.

[Rendered](https://github.com/benblank/rfcs/blob/exit-codes/text/0000-exit-codes.md)

### DIFF
--- a/text/0000-exit-codes.md
+++ b/text/0000-exit-codes.md
@@ -25,6 +25,10 @@ A list of exit codes must be maintained within the Notion source code for use by
 # Details
 [details]: #details
 
+## Windows exit codes
+
+This RFC focuses largely on compatibility with Unix exit codes because there appears to be little consensus as to exit code meaning on Windows, aside from `0` indicating success and other values indicating failure.  Certain system applications (such as the Windows installer) use [stardard Windows error codes](https://docs.microsoft.com/en-us/windows/desktop/Debug/system-error-codes) as exit codes, but that list is nearly fully packed in the 8-bit range and is excessively granular (as it extends well beyond the 8-bit range).  For these reasons, compatibility with standard error codes is not a goal of this document.
+
 ## Shims
 
 Of the three "first-class" shims (`node`, `npm`, and `yarn`), only Node currently publishes a [list of exit codes](https://nodejs.org/api/process.html#process_exit_codes); the other two only emit codes 0/1 for success/failure.  The exit codes for other shimmed executables is too large a space to examine, but Node follows what appears to be a common practice: simply numbering errors starting with 1, preserving the [special meanings commonly applied by POSIX shells](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_08_02).  This gives us incentive to avoid low-numbered exit codes (as likely to conflict) and exit codes with the high bit set (as typically representing termination due to signals).
--- a/text/0000-exit-codes.md
+++ b/text/0000-exit-codes.md
@@ -15,6 +15,8 @@ Each exit code is intended to signal to an external process some information abo
 
 Because redefining exit codes later in Notion's life would be a major hassle for third-party tools reliant on them, we should consider at the outset grouping errors together by exit code.
 
+Further, Notion must avoid conflicts between the exit codes produced by shims and those produced by the shimmed executables, in order to best preserve their behavior.
+
 # Pedagogy
 [pedagogy]: #pedagogy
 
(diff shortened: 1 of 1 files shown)


dherman:
I think the main thing I'd like to see here is separating out the following cases:
- exit codes from `notion`
- exit codes from `node`
- exit codes from `npm`
- exit codes from `yarn`
- exit codes from `npx` and third-party executables
I would divide them up into three categories:
1. `npm` and `yarn`: This is where I'm most interested in getting value from exit codes, because when integrating into build/CI scripts you really want to be able to detect the main categories of errors. (Especially the network error category.) We don't control those tools but maybe we can engage with those organizations and see if there are ways they could communicate structured output (either via standard exit codes or execution modes (enabled via command-line flag or env var) that produce parseable output?) so we have a better ability to capture and categorize failures?
2. `node`, `npx`, and third-party executables: We have no control over these whatsoever, but if we have Notion-specific errors we still need to produce _some_ sort of error code. How should we think about the fact that our exit codes and theirs have to share the same space of 255 values?
3. `notion`: We have total control here and can define whatever exit code scheme we want.

benblank:
WRT `npm` and `yarn`, I don't like the idea of having our shims "rewrite" the tools' exit codes.  Reaching out to those teams and trying to convince them of the value of good exit codes is another issue, but I think it's important that shims faithfully reproduce the behavior of the tools being shimmed, perhaps even to the point of producing a `0` exit code when the shimmed tool completes successfully even if Notion encounters errors (e.g. being unable to set up a new shim for a third-party executable after `yarn` installs it).  If we start rewriting exit codes, we open up a whole new can of worms and the potential for misleading existing scripts which rely on those tools' current behavior.

Let me do some thinking and I'll try to update the RFC later today.

benblank:
Apologies for the delay!  I finally had some time to work on this and have expanded and partially rewritten the RFC.

text/0000-exit-codes.md
```diff
@@ -0,0 +1,117 @@
...
+|       0 | Success
+|       1 | Unknown/"unfriendly" error
+|       2 | Unused (indicates builtin misuse in bash)
+|   3-125 | Errors
+| 126-127 | Unused (indicates failure to execute in POSIX shells)
+| 128-255 | Terminated due to signal `code - 128`
+
+## Implementation
+
+Exit codes shall be defined as an `enum` with appropriate documentation comments and an explicit integer value for each element.  Being explicit about values allows simple, at-a-glance correlation of exit codes to error types (as well as enabling sparse values).
```
dherman: This is a really, really great idea. :+1:

text/0000-exit-codes.md
```diff
@@ -0,0 +1,117 @@
...
+
+The primary downside introduced by this proposal is the loss of error granularity.  By grouping multiple errors into the same exit code, external tooling loses the ability to differentiate between those errors without capturing and parsing Notion's output.  However, it doesn't seem realistic to provide a separate exit code for each error due to the limited nature of the former and the limitless potential of the latter.  🙂
+
+Another issue is the communication of exit codes.  We want to publish the list as part of Notion's standard, user-facing documentation, but this requires either manually keeping the documentation in sync with the source or writing a tool to do so for us.  The alternative is to point interested parties directly to the Notion source, which isn't ideal.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- Is a process needed for 'vetting' new exit codes, to prevent duplication?  If so, is the RFC process appropriate?  Alternatively, is code review sufficient?
+- Should we engage npm/Yarn to improve error categorization (either through parsable output or diversifying exit codes)?
```
dherman: Yarn does have the ability to generate JSON formatted output, although currently you have to choose one or the other. I have reached out to some folks involved with Yarn to see if it'd be possible to "tee" the output via a config file or environment variable, so that a developer could choose whatever output they want at stdout/stderr, but we could specify a side file that also receives the output in JSON format.

text/0000-exit-codes.md
```diff
@@ -0,0 +1,117 @@
...
+[pedagogy]: #pedagogy
+
+A list of exit codes must be maintained within the Notion source code for use by developers, which should contain descriptions of when each code should be used.  This same list can be exposed via consumer-facing documentation to ensure that Notion developers and consumers share an understanding of the meaning of each exit code.
+
+# Details
+[details]: #details
+
+## Shims
+
+Of the three "first-class" shims (`node`, `npm`, and `yarn`), only Node currently publishes a [list of exit codes](https://nodejs.org/api/process.html#process_exit_codes); the other two only emit codes 0/1 for success/failure.  The exit codes for other shimmed executables is too large a space to examine, but Node follows what appears to be a common practice: simply numbering errors starting with 1, preserving the [special meanings commonly applied by POSIX shells](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_08_02).  This gives us incentive to avoid low-numbered exit codes (as likely to conflict) and exit codes with the high bit set (as typically representing termination due to signals).
```
dherman: Don't forget `npx`, too.

dherman:
Oh, one more thing: it looks like you've done (very thorough!) research on Unix/POSIX signals, but what about Windows conventions for exit codes?

text/0000-exit-codes.md
```diff
@@ -0,0 +1,117 @@
...
+
+Additionally, shims must do their utmost to preserve the exit code supplied by the shimmed executable, only introducing a shim-specific exit code if the executable cannot be run at all.  To this end, shims must return 0 when the shimmed executable does, even if the shim itself is not able to perform its work (for example, if a new shim cannot be created when installing a new executable via `npm` or `yarn`).
```
dherman: Really insightful point. With the idea of pushing harder on statelessness, some of these tricky edge cases may get simpler—for example, I think in the end we won't actually need `npm i -g`/`yarn global add`/etc to create shims. Generally, a more stateless model has fewer opportunities to fail, since it doesn't make as many persistent changes to the filesystem.

text/0000-exit-codes.md
benblank: Is `npx` a first-class shim (i.e. does it have special handling attached to it)?  I had thought it was being handled as a "generic" 3rd-party shim.

benblank:
> Oh, one more thing: it looks like you've done (very thorough!) research on Unix/POSIX signals, but what about Windows conventions for exit codes?

I added a section addressing Windows exit codes and why I'm not proposing compatibility with them.

dherman:
This is looking great and looks ready to merge to me. If no one raises any issues in the next couple days I'll merge it.

PR #21, opened 2018-08-21
Package executables

mikrostew:
> Many Node packages contain one or more executable files that they would like to install into the PATH. Notion should make these executables available to the user:
>
> - locally, on a per-project basis (project tools)
> - globally, installed as if they are static binaries (user tools)

[Rendered](https://github.com/mikrostew/rfcs/blob/6e218eae3b5c87f7ca14b2d5b81fcc0ae788dfdf/text/0000-package-executables.md)

### DIFF
--- a/text/0000-package-executables.md
+++ b/text/0000-package-executables.md
@@ -1,0 +1,177 @@
+- Feature Name: package_executables
+- Start Date: 2018-08-20
+- RFC PR: (leave this empty)
+- Notion Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Many Node packages contain one or more executable files that they would like to install into the PATH. Notion should make these executables available to the user:
+
+- locally on a per-project basis (project tools)
+- globally, installed as if they are static binaries (user tools)
+
+# Motivation
+[motivation]: #motivation
+
+## Project Tools
+
+Notion will provide some killer features for working in Node projects (that no other version managers have):
+
+### Isolating the project scripts:
+
+As it currently works, any package executables referenced in the `"scripts"` resolve to executables installed locally for the project. But, these scripts can also make use of global executables, which may not be consistent for different collaborators. Notion removes this loophole by stripping the global package binaries from the PATH so the scripts run as if there are no global executables installed. This provides consistent environments for project collaborators as well as CI.
+
+### Command-line access to package executables:
+
+When the user is in the Node project, Notion gives command-line access to the package executables for the project. This is normally only possible through the project scripts, but that requires 1) setting up the script, and 2) running `npm run <script>`. With Notion, the user can just run the package executable directly, like `ember build`, instead of having to setup a script like `"build": { "ember build" }` and then `npm run build`.
+
+
+## User Tools
+
+Once a user installs a package executable globally, it should just work, and continue to work. The availability of the executable should not depend on the active or installed versions of Node.
+
+Most version managers provide the ability to install package executables on a per-version basis. Once a user switches Node versions, they must re-install the executable for it to be available. Notion will eliminate this hassle - once the executable is installed, it will be statically linked to the Node version used to install it, and it will always run with a pristine version of Node. It will not break because of version switches, and its environment will not be polluted by other globally-installed package executables.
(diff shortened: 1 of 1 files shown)


rwjblue:
Overall this is a really nice write up, and helps rationalize the mental model around these tools. 

I’m still struggling a bit on the `shims` section though. It seems like the shims are only used to defer to project local executables, right? Since the shims are globally installed (AFAICT there is only one set of them in the `~/.notion/.bin` directory), the `$PATH` (along with tools like `which some-name`) still expose the executable name in areas where it is actually not supported.

rwjblue:
> Now, as you say, this does show through when you do things like which some-name. We also need a notion which command to find out the actual command that the shim will run, so you can see past the shims.

Hehe, well, we could allow folks to shim `which` itself (probably should be an opt-in?)...

dherman:
@rwjblue We'd definitely discussed that, but it makes me nervous. It feels like it's overriding a primitive of the shell that you can't easily get back. And it feels like it's asking for bugs, for example if a shell script saves `SOME_TOOL=$(which some-tool)` in one directory and then changes directories and calls `$($SOME_TOOL)`, it'll basically circumvent the behavior of Notion, even though the user had Notion installed.

I think another way to think of this is that `which` is a primitive of the shell, and Notion is operating at a higher layer of abstraction than `which`. That's a bit hand-wavy and philosophical, but I suspect in practice it may work out ok. Usage experience will give us a sense.

rwjblue:
Right, that makes sense. I agree that it definitely is quite "magical", and perhaps just being up front with what is going on is better. Either way it is certainly a better starting point.

text/0000-package-executables.md
```diff
@@ -0,0 +1,177 @@
...
+Any package executables installed by direct `dependencies` and `devDependencies` of a Node project are available for the user to use when they are in that project. These executables will run using the Node version that is configured for the project.
+
+### User tools
+
+When a user installs a package executable globally, it’s like they are installing a static binary. It will work in the same way that it was installed, despite changing directories or changing Node versions.
+
+# Details
+[details]: #details
+
+To make this work, Notion will put shims on the PATH in `~/.notion/shims/` to intercept execution of package executables and delegate appropriately. These shims are implemented as symlinks to the `launchbin` executable, where the name of the symlink is the name of the shimmed executable. Since the shims directory will come first (or very early) in the PATH, when any shimmed executable is run, it will execute `launchbin` with the name of the executable as the first argument.
```
dherman: It's `~/.notion/bin/` now.

dherman: It won't actually be guaranteed to pass the name of the executable as arg0. Thankfully we don't need to rely on that, since it looks like there are [cross-platform mechanisms for finding out the current executable](https://docs.rs/proc_self/0.1.0/proc_self/fn.exe_path.html).

text/0000-package-executables.md
```diff
@@ -0,0 +1,177 @@
...
+
+Then `launchbin` uses this logic to determine how to delegate execution:
+
+- If the user is in a Node project, and the executable is a direct dependency of the current project, then that is executed (using the absolute path to the executable). We determine that an executable is a direct dependency by reading the `"bin"` section of each direct dependency’s `package.json`. Only direct dependencies are executed because `npm` and `yarn` install executables for the complete tree of project dependencies, not just direct dependencies. Exposing all of those would be a security risk as well as confusing to debug.
+- Otherwise this is installed as a user tool, so that is executed (using the absolute path to the executable).
```
dherman: This case should _check_ if it's installed as a user tool. Then there's a subsequent case, which is to delegate to whatever system tool might be there in the remainder of `PATH`.

dherman:
It occurs to me that I kind of bigfooted this whole RFC with the reproducibility RFC. Or at least, they both have some overlap.

And at the same time, as I commented on that one, I want to remove all the git stuff from it, which is a relatively significant portion of it.

Maybe the right thing to do is reject both RFCs and consolidate them into one new one.

dherman:
Closing since this is superseded by #27.

PR #22, opened 2018-09-14
RFC: reproducibility

dherman:
From the RFC:

> This RFC proposes a toolchain management model for Notion that’s built around a primary goal of *reproducibility*: that is, giving users end-to-end control around toolchains that, once set up, always exhibit identical behavior. The resulting user experience should be **”install and forget”** — after setting up a toolchain, a user or team never has to worry about the many forms of *tool bitrot*—hidden state changes (e.g. the addition or removal of global packages), subtle differences in tool versions—on an existing machine, between project collaborators’ machines, or when setting up a new machine. In short, reproducibility means Notion’s declarative configuration completely determines a Node environment’s behavior.

[Rendered](https://github.com/notion-cli/rfcs/blob/d6ba0b7e62b346bc2af0125ce3c19117ac2cc237/text/0000-reproducibility.md)

### DIFF
--- a/text/0000-reproducibility.md
+++ b/text/0000-reproducibility.md
@@ -1,0 +1,257 @@
+- Feature Name: reproducibility
+- Start Date: 2018-09-14
+- RFC PR: 
+- Notion Issue: 
+
+# Summary
+[summary]: #summary
+
+This RFC proposes a toolchain management model for Notion that’s built around a primary goal of *reproducibility*: that is, giving users end-to-end control around toolchains that, once set up, always exhibit identical behavior. The resulting user experience should be **”install and forget”** — after setting up a toolchain, a user or team never has to worry about the many forms of *tool bitrot*—hidden state changes (e.g. the addition or removal of global packages), subtle differences in tool versions—on an existing machine, between project collaborators’ machines, or when setting up a new machine. In short, reproducibility means Notion’s declarative configuration completely determines a Node environment’s behavior.
+
+# Motivation
+[motivation]: #motivation
+
+One important motivation of this RFC is to reconcile two competing constraints in the Node ecosystem:
+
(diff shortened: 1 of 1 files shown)


text/0000-reproducibility.md
```diff
@@ -0,0 +1,257 @@
...
+
+Another reasonable criticism is that pinning the Node version for tools in the user toolchain means that users will not automatically benefit from performance and security improvements in Node. There are a couple of reasons this is outweighed by the benefits of pinning. First, as users upgrade the tool version itself, they will automatically get re-pinned to the newer version of Node specified by newer versions of the tool’s `package.json`. Second, we should at least allow users the option to override the tool’s specified Node version. But by letting the tool choose its platform version by default, user’s have a stronger guarantee that the tool will work and continue to work consistently based on how it was tested.
```
mikrostew: This RFC is missing a discussion about installing user tools that do not have a pinned node version. Currently there are 0 tools that specify a `"toolchain"`, and even if Notion is wildly popular there will be a transition period where some tools have a pinned `node` version and some do not. Actually there may always be some tools that do not specify a `"toolchain"`, for whatever reason.

Notion should be able to handle this situation and preserve reproducibility. That could be using the version of node in the user's config, or if that is not specified, using the user's default version of node, and pinning that somehow. In these cases Notion could modify the installed tool's package.json to include a `"toolchain"` section if it is not already present.

dherman:
I think what I need to do next is rip out the discussion of platform packs and git representations, which should be separately investigated as an optimization technique down the road. That way this RFC can stay focused on the core ideas of reproducibility, which are mostly orthogonal to that implementation approach.

dherman:
Closing since this is superseded by #27.

PR #24, opened 2018-10-16
RFC: `notion pin`

dherman:
Summary:

> This RFC proposes renaming the `notion use` command to `notion pin`.

[Rendered](https://github.com/dherman/rfcs-2/blob/pin/text/0000-notion-pin.md)

### DIFF
--- a/text/0000-notion-pin.md
+++ b/text/0000-notion-pin.md
@@ -6,7 +6,7 @@
 # Summary
 [summary]: #summary
 
-One paragraph explanation of the feature.
+This RFC proposes renaming the `notion use` command to `notion pin`.
 
 # Motivation
 [motivation]: #motivation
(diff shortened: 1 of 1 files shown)


rwjblue:
As a long time `nvm` user I was pretty confused when initially reading docs and repo source until I realized that `notion use` was very nearly the opposite of what I thought it was. Obviously all future users of Notion won’t have had the same baggage, but if we can solve that confusion **and** have a better name for new users all the better.

To me, `notion pin` threads the needle nicely here. It certainly seems more intuitive...

dherman:
I'm marking this RFC as being in final comment period! I'll announce to the channel.

charlespierce:
LGTM! I think `pin` does a great job of being an intuitive command

PR #25, opened 2018-11-01
RFC: Config -> Hooks

charlespierce:
> Rename `Config` to `Hooks` and allow additional methods for specifying tool resolution.

[Rendered](https://github.com/charlespierce/rfcs/blob/master/text/0000-config-hooks.md)

### DIFF
--- a/text/0000-config-hooks.md
+++ b/text/0000-config-hooks.md
@@ -71,3 +71,4 @@ There are some performance concerns with calling an external process to resolve 
 [unresolved]: #unresolved-questions
 
 - Will an organization need or want to manage the individual tool settings in separate files?
+- Should we change the format of the Hooks file(s) to JSON, since the end-users are JS devs who will be more familiar with JSON than TOML?
--- a/text/0000-config-hooks.md
+++ b/text/0000-config-hooks.md
@@ -1,0 +1,73 @@
+- Feature Name: config_hooks
+- Start Date: 10/31/2018
+- RFC PR: (leave this empty)
+- Notion Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Rename `Config` to `Hooks` and allow additional methods for specifying tool resolution.
+
+# Motivation
+[motivation]: #motivation
+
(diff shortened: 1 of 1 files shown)


text/0000-config-hooks.md
```diff
@@ -0,0 +1,73 @@
...
+# Details
+[details]: #details
+
+## Rename
+
+This change is relatively straightforward, renaming from `Config` to `Hooks` internally, as well as in any documentation. It should also include switching the file from which these settings are loaded from `config.toml` to `hooks.toml`.
+
+## URL Prefix Resolver
+
+This option should store the user-defined URL prefix and then use the default behavior for determining the variable part of the URL in any resolution.
```
mikrostew: Will we need to support multiple URL prefixes for this case, depending on how the repository is organized?

charlespierce: My thinking was that the URL Prefix would be the more simplistic option for cases where it works. Each of the hooks would be configurable per-tool and per-use (one use being the tool version index and the other being the binary file resolution path), so we would support different URL Prefixes for different tools and uses, but for a given tool use case only a single prefix would be supported.

If they need more involved paths then that can be supported with the URL Template hook.

charlespierce:
Another open question that just came to me: Should the hooks file(s) be `.toml`, or `.json` files? While `.toml` is an exceedingly simple format, the end-users of the tool ultimately are JS devs. They will already be familiar with JSON, so it might reduce friction to have the hooks / file-based configuration be in JSON format.

dherman:
@charlespierce Fair question Re: TOML; the unfamiliarity is an onboarding tax for JS devs. OTOH, comments! :P

I think we should stick with TOML for now and wait to see how loudly people complain. Configuring hooks isn't something people will typically do in normal workflows anyway.

dherman:
This looks great! I'm going to mark it as being in final comment period and announce it to the channel.

rwjblue:
> I think we should stick with TOML for now and wait to see how loudly people complain. Configuring hooks isn't something people will typically do in normal workflows anyway.

I tend to agree here.

chriskrycho:
This has been in FCP for just shy of three months. Should it be marked as accepted and merged… especially since the implementation is already merged? 😁

PR #30, opened 2019-02-26
RFC: migrate from Docopt to StructOpt

chriskrycho:
Use [StructOpt] to get automatic coloration, completions, and (in the author's experience) easier development and maintenance.

[StructOpt]: https://docs.rs/structopt/0.2.14/structopt/

[Rendered](https://github.com/chriskrycho/notion-rfcs/blob/cli-ui/text/0000-cli-ui.md)

### DIFF
--- a/text/0000-cli-ui.md
+++ b/text/0000-cli-ui.md
@@ -32,7 +32,7 @@ Switching to [StructOpt] improves our story in each of these areas:
 
 - It properly passes through global options to subcommands, and it understands all variants of “help” invocations (position and both flag styles) out of the box, and can be trivially configured to display help/usage information when no arguments are passed.
 
-- Colorized help output works out of the box as a setting which can be enabled globally as well as on a per-command basis.
+- Cross-platform olorized help output works out of the box as a setting which can be enabled globally as well as on a per-command basis.
 
 Though StructOpt is not *universally* better than Docopt, it can generate similar output, but with a lower maintenance burden and a number of niceties *not* supported by Docopt.
 
--- a/text/0000-cli-ui.md
+++ b/text/0000-cli-ui.md
@@ -35,7 +35,9 @@ Though StructOpt is not *universally* better than Docopt, it can generate simila
 # Pedagogy
 [pedagogy]: #pedagogy
 
-This is purely an internal, developer-facing change, with *benefits* to the user but no changes in actual use of Notion. The documentation for StructOpt (and its core dependency [`clap`]) are sufficient for developers working on the project.
+This change improves the user experience but does not change the user workflow in any way.
+
+It does change the *developer* workflow; however, no additional internal documentation is needed. The documentation for StructOpt (and its core dependency [`clap`]) are sufficient for developers working on the project.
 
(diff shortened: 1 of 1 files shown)


chriskrycho:
Happily, no, because we won't actually be using the `template` field at this point – if it ends up being useable for us later, it would be if we made e.g. `ansi_term` support `const fn` (if that's possible; it may not be!) and thus got its cross-platform support baked in for free. The basic colorization stuff all Just Works cross-platform out of the box.

text/0000-cli-ui.md
```diff
@@ -0,0 +1,96 @@
...
+- The configuration is validated by the compiler rather than at runtime, and the compiler’s error messages are usually much clearer about what is wrong than are the runtime errors from Doctopt.
+
+- It properly passes through global options to subcommands, and it understands all variants of “help” invocations (position and both flag styles) out of the box, and can be trivially configured to display help/usage information when no arguments are passed.
+
+Though StructOpt is not *universally* better than Docopt, it can generate similar output, but with a lower maintenance burden and a number of niceties *not* supported by Docopt.
+
+# Pedagogy
+[pedagogy]: #pedagogy
+
+This is purely an internal, developer-facing change, with *benefits* to the user but no changes in actual use of Notion. The documentation for StructOpt (and its core dependency [`clap`]) are sufficient for developers working on the project.
```
dherman: I find this sentence confusing: it sounds like it's saying "this is user-visible but it's invisible to the user." Do you mean it does have user benefits but doesn't have any changes to workflow?

chriskrycho: Ah, great point. I'll update to that effect; thank you!

text/0000-cli-ui.md
```diff
@@ -0,0 +1,96 @@
+- Feature Name: cli-ui-improvements
+- Start Date: 2019-02-25
+- RFC PR: (leave this empty)
+- Notion Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Transition from [Docopt] to [StructOpt], for an easier path to nice user experience (and better developer experience when working on) the command line interface.
```
dherman: Are completions the only user-visible change? Is there an implication for color? You mention it later but it's not clear to me if DocOpt doesn't support color and StructOpt does?

dherman:
I'm 👍 on this. I like DocOpt but agree that the runtime parsing, unhelpful errors, and lack of documentation around the input grammar are problematic. And generally I think it's not super actively maintained so that's probably another good reason to switch.

text/0000-cli-ui.md
chriskrycho: Ah, good clarification. Docopt *can* support them, but you have to manually do every bit of it with e.g. `format!` and `ansi_term`; they work out of the box (and are pretty nice!) with StructOpt.

chriskrycho:
Thanks, @dherman and @charlespierce – I've updated the RFC to clarify those points!

dherman:
Discussion on this RFC has stabilized and the implementation has already landed and is working out. I'm marking this as final comment period and we can probably merge soon.

PR #31, opened 2019-02-28
RFC: Helpful Error Messages

charlespierce:
>Improve error messages so that users are better able to understand and troubleshoot what went wrong.

[Rendered](https://github.com/charlespierce/rfcs/blob/error_messages/text/0000-helpful-error-messages.md)


text/0000-helpful-error-messages.md
```diff
@@ -0,0 +1,68 @@
...
+
+# Critique
+[critique]: #critique
+
+The primary point of critique is with the exact voice and style we want to use for the error messages. There are two extremes we could take, though this proposal is for doing a middle-ground that hopefully gets the benefits of each without significant drawbacks.
+
+On the one hand, we could argue that since Notion is a CLI tool, and many of the commands (e.g. `npm run ...`) will be run by automated scripts, we should keep the error messages as concise as possible. This prevents up from filling up logs with human-readable text, while still providing feedback in the form of an error code and short description. The issue with this is that Notion will also be used very often in non-automated contexts. Putting an extra step between the user and useful feedback hurts the user experience and makes it less likely for the helpful information to actually be of use.
```
rwjblue:
```suggestion
On the one hand, we could argue that since Notion is a CLI tool, and many of the commands (e.g. `npm run ...`) will be run by automated scripts, we should keep the error messages as concise as possible. This prevents us from filling up logs with human-readable text, while still providing feedback in the form of an error code and short description. The issue with this is that Notion will also be used very often in non-automated contexts. Putting an extra step between the user and useful feedback hurts the user experience and makes it less likely for the helpful information to actually be of use.
```

dherman:
Some questions worth addressing:

- What should the approach to line lengths be? Should we have a fixed line length, or hard wrap based on the current console's dimensions?
- What sort of prefix should we use? e.g.:
  * `error: `
  * `Error: `
  * `Notion error: `
  * etc
  * Note that it's important to include "Notion" for shim errors, to help the user understand that the error happened at a different context than the one that was probably in their mind at the time they were trying to execute the shim. For the `notion` tool itself, including "Notion" in the prefix might look redundant.
- How should we use color/styling? I'd suggest it's helpful at least for the prefix to be red and bold, because it helps provide a more immediate signal to the user than the text itself, and also helps for scanning visually through a longer sequence of console output.
- Are there places where we it would be valuable for us to show context about an error? Notion isn't the same kind of product as a compiler, so many errors are more about e.g. filesystem issues. But maybe for things like invalid configuration files, it would be helpful to show some of the context surrounding the invalid file contents. This might be lower priority to implement than some of the other error improvements, but it might be good to call out in the RFC as a possible improvement.

Especially when considering some of these visual questions, some screenshots could be helpful to make the RFC more concrete.

Also, let me suggest one strategy for making progress on the prose style: you could do a first pass through the `ErrorDetails` enum, grabbing a set of maybe 5 - 10 errors that cover a reasonably diverse range of different kinds of errors, and try to develop specific error messages for each and include them in this RFC. That would help start to establish a prose style, and might help us to divide up the work on the remaining cases.

charlespierce:
Updated to cover the style questions. I like the idea of tackling a few selected errors and taking screenshots to help illustrate the new messages, I'll work on that this week.

chriskrycho:
> Should we have a fixed line length, or hard wrap based on the current console's dimensions?

Using [textwrap](https://docs.rs/textwrap/0.11.0/textwrap/index.html) (and probably others; that's just the one I've played with) you can fairly easily set a max width of e.g. 72 characters for legibility, but wrap if the terminal is narrower than that.

rwjblue:
Seems like this has landed already (https://github.com/volta-cli/volta/pull/344, https://github.com/volta-cli/volta/pull/340, https://github.com/volta-cli/volta/pull/342, https://github.com/volta-cli/volta/pull/338, https://github.com/volta-cli/volta/pull/339, https://github.com/volta-cli/volta/pull/337, https://github.com/volta-cli/volta/pull/336, https://github.com/volta-cli/volta/pull/326, https://github.com/volta-cli/volta/pull/328), any reason not to merge the RFC?

dherman:
@rwjblue ya, should've been merged already. Doing it now!

PR #32, opened 2019-04-24
RFC: update version handling.

chriskrycho:
[Rendered](https://github.com/chriskrycho/notion-rfcs/blob/update_version_handling/text/0000-update-version-handling.md)

TL;DR: support `notion install yarn@latest node@10.5 typescript@3 create-react-app`.

### DIFF
--- a/text/0000-update-version-handling.md
+++ b/text/0000-update-version-handling.md
@@ -1,0 +1,225 @@
+- Feature Name: update_version_handling
+- Start Date: 2019-04-23
+- RFC PR: (leave this empty)
+- Notion Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Change all version-specifying commands to use the `<tool>[@<version>]` syntax from `npm install` and `yarn add`.
+
+# Motivation
+[motivation]: #motivation
+
+Users may regularly want to add multiple tools to their toolchain at the same time. They are also accustomed to being able to install multiple dependencies simultaneously with the `npm install` or `yarn add` commands, both in a package and globally.
+
+- **Local installs:**
+
+    ```sh
+    npm install ember-cli@3.9.1 typescript@3
+    ```
+    
+    ```sh
+    yarn global ember-cli@3.9.1 typescript@3
+    ```
+
+- **Global installs:**
+
+    ```sh
+    npm install --global ember-cli@3.9.1 typescript@3
+    ```
+    
+    ```sh
+    yarn global add ember-cli@3.9.1 typescript@3
+    ```
+
+Notion currently does not support this, requiring individual invocations for each tool:
+
+- **User toolchain:**
+
+    ```sh
+    notion install ember-cli 3.9.1
+    notion install typescript 3
+    ```
+
+- **Project toolchain:**
+
+    ```sh
+    notion pin ember-cli 3.9.1
+    notion pin typescript 3
+    ```
+
+- **Tool retrieval:**
+
+    ```sh
+    notion fetch ember-cli 3.9.1
+    notion fetch typescript 3
+    ```
+
+This is a surprising difference for users coming from `npm` or `yarn`. Switching over to use the same invocation as the existing tools will make adoption smoother and less surprising.
+
+# Pedagogy
+[pedagogy]: #pedagogy
+
+Installation, fetching, and pinning function much as they do today, with two changes:
+
+- The tool + version combination is now a single string, in the format `tool[@<version>]`. That is, a user may supply a tool with its name only, e.g. `node`, or with a specific version using the `@` to separate the name from the version, e.g. `node@10`, `node@10.5`, `node@10.5.2`, or `node@latest`.
+
+- The invocations now allow multiple tool + version combinations, each of the format specified above. For example, a user may simultaneously pin a Node and a Yarn version (with any combination of the version specifier, or none at all):
+
+  ```sh
+  notion install node@10.5 yarn
+  ```
+
+## Explanation for Users
+
+### Current Node users
+
+There are two broad sets of current Node users: those not using a Node version management tool, and those using an existing tool like `nvm`, `nodenv`, or `nvm-windows`.
+
+Current Node users who are not using a Node version management tool should find this a simplification: it matches what npm and Yarn do.
+
+Current Node users using existing tools will find this to be different from their experience:
+
+- `nvm` uses `nodenv install <version>`
+- `nodenv` uses `nodenv install <version>`
+- `nvm-windows` uses `nodenv install <version> [arch]` (where `[arch]` allows users to specify 32-bit or 64-bit builds)
+
+Although we are breaking with these tools, all of them are responsible to manage *only* the Node version, *not* other tools. Notion's purview extends to other tools and therefore overlaps more with the behavior expected from invoking `npm` or `yarn`.
+
+### New Node users
+
+New Node users may have a wide variety of different expectations for adding a version of a tool:
+
+- RubyGems: `gem install <gem name> [-v|--version <version>]`
+
+- Bundler: `bundle add <gem name> [--version=<version>]`
+
+- Nuget: `nuget install <package> [-Version <version>]`
+
+- Paket: `paket add <package> [--version <version spec>]`
+
+- `dotnet`: `dotnet tool install <PACKAGE_NAME> <-g|--global> [--version]`
+
+- Cargo: `cargo install crate [--version <version>]`
+
+- `cargo-edit`: `cargo add <package>[@<version>]...`
+
+Of these, only `cargo-edit` allows adding multiple packages simultaneously; it does so with the same interface as `npm` and `yarn` and proposed here.
+
+# Details
+[details]: #details
+
+## Installing, fetchin, or pinning a single tool
+
+Managing a tool with its default version is unchanged.
+
+| Original                  | Proposed                  |
+| ------------------------- | ------------------------- |
+| `notion install node`     | `notion install node`     |
+| `notion pin yarn`         | `notion pin yarn`         |
+| `notion fetch typescript` | `notion fetch typescript` |
+
+Managing a tool at a specific version, at any level of granularity, now uses `@` to separate the tool name from its version.
+
+| Original                     | Proposed                     |
+| ---------------------------- | ---------------------------- |
+| `notion install node latest` | `notion install node@latest` |
+| `notion pin node 10`         | `notion pin node@10`         |
+| `notion fetch node 10.5`     | `notion fetch node@10.5`     |
+
+This also works for shorthands (currently `latest`, conceivably also tags like `next` or `lts` in the future):
+
+| Original                 | Proposed                  | Status       |
+| ------------------------ | ------------------------ | ------------ |
+| `notion pin node latest` | `notion pin node@latest` | current      |
+| `notion pin node next`   | `notion pin node@next`   | hypothetical |
+| `notion pin node lts`    | `notion pin node@lts`    | hypothetical |
+
+## Installing, fetching, or pinning multiple tools
+
+Previously, installing multiple tools required invoking Notion multiple times. Now, it may be invoked with multiple tools at once.
+
+- **Original:**
+
+    ```ts
+    notion install create-react-app latest
+    notion install typescript 3.4
+    ```
+
+- **Updated:**
+
+    ```
+    notion install create-react-app@latest typescript@3.4
+    ```
+
+There are no impacts to features other than fetching, installing, and pinning.
+
+## Implementation changes
+
+The command line implementation will need to be updated, with the install, fetch, and pin commands now receiving a list of tool specs.
+
+The command line parsing will need to be updated to accept variadic arguments of tool-with-optional-versions, using Clap's [`Arg::multiple`](https://docs.rs/clap/2.32.0/clap/struct.Arg.html#method.multiple) argument, with a custom parser. E.g., for `install`, we might have something like this
+
+```rust
+#[derive(StructOpt)]
+pub(crate) struct Install {
+    /// The tool to install, e.g. `node` or `npm` or `yarn`
+    #[structopt(multiple = true), parse(try_from_str = "ToolSpec::try_from_str")]
+    tools: Vec<ToolSpec>,
+}
+```
+
+Here we assume that the `ToolSpec` will add a `try_from_str` or `try_from_os_str` implementation, with a useful error type—here hand-waved as `String`, but it could be as complex as we need as long as the error type has an implementation of `Into<String>`:
+
+```rust
+impl ToolSpec {
+  fn try_from_str(string: &str) -> Result<ToolSpec, String> {
+    // split apart the tool name and its version
+  }
+}
+```
+
+This is a relatively simple implementation. The only complication is the existence of Node namespaces: `@namespace/package`. This means parsing the version is not as simple as splitting on `"@"`. However, it should still be straightforward using the combination of a regular expression and the existing semver tooling.
+
+The `ToolSpec` change will make the format `tool[@version]` work; updating the commands to take a `Vec<ToolSpec>` will support passing multiple arguments.
+
+# Critique
+[critique]: #critique
+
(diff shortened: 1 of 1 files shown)


text/0000-update-version-handling.md
```diff
@@ -0,0 +1,225 @@
...
+    ```
+
+This is a surprising difference for users coming from `npm` or `yarn`. Switching over to use the same invocation as the existing tools will make adoption smoother and less surprising.
+
+# Pedagogy
+[pedagogy]: #pedagogy
+
+Installation, fetching, and pinning function much as they do today, with two changes:
+
+- The tool + version combination is now a single string, in the format `tool[@<version>]`. That is, a user may supply a tool with its name only, e.g. `node`, or with a specific version using the `@` to separate the name from the version, e.g. `node@10`, `node@10.5`, `node@10.5.2`, or `node@latest`.
```
mikrostew: This RFC uses "version", but would it be more accurate to say "version specifier", or maybe "version range"?

These invocations are mentioned in the doc:

```
node
node@latest
node@1.2.3
node@1.2
```

It is my intuition that it is also possible to do these kinds of things, which are supported by `npm install` and `yarn add`:

```
node@10.1*
node@^10.1
node@~10.1
node@>=10
node@<10
```

These operations are already supported by `jetson install <package> <versionspec>` (through the use of the `semver` crate), but it may be worth spelling out more concretely what is and is not supported here.

For instance there are also other range operators like `1.2.3 - 3.4.5`, `1.2 || 2.2`, etc. that _could_ be supported in the existing `notion install <package> <versionspec>` case, but would probably be too complicated to parse for this change.

text/0000-update-version-handling.md
```diff
@@ -0,0 +1,225 @@
...
+    Does it refer to the latest version of `create-react-app`, or the default version of `create-react-app` and [the `latest` package](https://www.npmjs.com/package/latest)? For all non-SemVer-string parses, this will be impossible to resolve.
+
+    The proposed design here is well-proven with other tools which install more than simply a single tool-and-version pair, e.g. `npm`, `yarn`, and `cargo-edit`, and it does not fall prey to these problems of indeterminacy.
+
+-   Leave things as they are. This has the upside of matching other Node version managers, all of which use a close variant on our current syntax.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- This will *suggest* to users that they can use tags, e.g. `tool@next`, a feature we do not currently support. Should we include support for that here, or plan it as follow-on work, or neither?
```
mikrostew: I think this should be planned as separate work. This is something we could already implement separate from the changes here, and this RFC would get the benefit of that. (Assuming the changes in this RFC will re-use most of the existing code to parse the input version from the command line)

text/0000-update-version-handling.md
```diff
@@ -0,0 +1,225 @@
...
+
+### Current Node users
+
+There are two broad sets of current Node users: those not using a Node version management tool, and those using an existing tool like `nvm`, `nodenv`, or `nvm-windows`.
+
+Current Node users who are not using a Node version management tool should find this a simplification: it matches what npm and Yarn do.
+
+Current Node users using existing tools will find this to be different from their experience:
+
+- `nvm` uses `nodenv install <version>`
```
dherman: copy-pasta bug -- `nvm install <version>`

text/0000-update-version-handling.md
```diff
@@ -0,0 +1,225 @@
...
+- `nodenv` uses `nodenv install <version>`
+- `nvm-windows` uses `nodenv install <version> [arch]` (where `[arch]` allows users to specify 32-bit or 64-bit builds)
```
dherman: copy-pasta bug -- `nvm install <version> [arch]`

text/0000-update-version-handling.md
```diff
@@ -0,0 +1,225 @@
...
+  notion install node@10.5 yarn
+  ```
+
+## Explanation for Users
+
+### Current Node users
+
+There are two broad sets of current Node users: those not using a Node version management tool, and those using an existing tool like `nvm`, `nodenv`, or `nvm-windows`.
+
+Current Node users who are not using a Node version management tool should find this a simplification: it matches what npm and Yarn do.
```
dherman: And also npx!

text/0000-update-version-handling.md
chriskrycho: Doh! Thanks! Pushing for this and the dupe below!

text/0000-update-version-handling.md
chriskrycho: Ah, yes I never got in the npx habit so I always forget it!

text/0000-update-version-handling.md
chriskrycho: This is a great call-out – and when I started poking at the implementation, I realized that "version" and "version spec" are different. I'll update it concretely to specify what we *don't* support here as well. I think we'll support versions and version specs, but not range operators (I'll add those to the <b>Unresolved Questions</b> as a thing I think we should defer – probably permanently).

text/0000-update-version-handling.md
chriskrycho: That's my take as well, but I wanted to raise it. I also learned in poking through the implementation that we *do* have `lts` support, so I'll make sure the reference to that in the table is updated!

text/0000-update-version-handling.md
```diff
@@ -0,0 +1,225 @@
...
+- Cargo: `cargo install crate [--version <version>]`
+
+- `cargo-edit`: `cargo add <package>[@<version>]...`
+
+Of these, only `cargo-edit` allows adding multiple packages simultaneously; it does so with the same interface as `npm` and `yarn` and proposed here.
+
+# Details
+[details]: #details
+
+## Installing, fetchin, or pinning a single tool
```
charlespierce: Minor typo: `fetching`
```suggestion
## Installing, fetching, or pinning a single tool
```

text/0000-update-version-handling.md
```diff
@@ -0,0 +1,225 @@
...
+
+Managing a tool at a specific version, at any level of granularity, now uses `@` to separate the tool name from its version.
+
+| Original                     | Proposed                     |
+| ---------------------------- | ---------------------------- |
+| `notion install node latest` | `notion install node@latest` |
+| `notion pin node 10`         | `notion pin node@10`         |
+| `notion fetch node 10.5`     | `notion fetch node@10.5`     |
+
+This also works for shorthands (currently `latest`, conceivably also tags like `next` or `lts` in the future):
```
charlespierce: `lts` is actually already supported (and is actually the default for `notion install node` now), as of this PR: https://github.com/notion-cli/notion/pull/331

text/0000-update-version-handling.md
charlespierce: Agreed. We'll also need to figure out how to handle tags on the `node` version, which doesn't have the same npm registry concepts that other tools have, since it's a level above `npm`.

text/0000-update-version-handling.md
charlespierce: Interestingly, in the current implementation, we actually do support all of the operators, including the range and `||` operators. The only trick is that it has to be parsed as a single argument, so you need to either quote it or remove all the whitespace. But I just tried `notion install node "8 || 9"` and it correctly found and installed Node 9.

text/0000-update-version-handling.md
```diff
@@ -0,0 +1,225 @@
...
+```
+
+This is a relatively simple implementation. The only complication is the existence of Node namespaces: `@namespace/package`. This means parsing the version is not as simple as splitting on `"@"`. However, it should still be straightforward using the combination of a regular expression and the existing semver tooling.
+
+The `ToolSpec` change will make the format `tool[@version]` work; updating the commands to take a `Vec<ToolSpec>` will support passing multiple arguments.
+
+# Critique
+[critique]: #critique
+
+This introduces breaking changes—something which should not be done lightly! It does so under the assumption that fetching or especially installing or pinning multiple tools simultaneously is a relatively common operation. This assumption may be wrong. It may also be the case that implementing this—with the required changes to
```
charlespierce: I suspect that installing / pinning tools is actually not going to be a common operation, since they are mostly done once per user / project and then not touched again until an upgrade is necessary. That said, I think that the benefit to the initial onboarding and bootstrapping of new projects is worth it. Being able to get your environment set up with a single command: `notion install node yarn ember-cli typescript` instead of 4 separate commands is a tangible benefit.

text/0000-update-version-handling.md
```diff
@@ -0,0 +1,225 @@
...
+    notion pin create-react-app
+    notion pin create-react-app@3.0
+    notion pin create-react-app 3.0
+    notion pin create-react-app typescript
+    notion pin create-react-app@3.0 typescript
+    notion pin create-react-app typescript@3.4
+    notion pin create-react-app@3.0 typescript@3.4
+    ```
+
+    However, this introduces substantial additional complexity for the implementation, and ambiguity around certain invocations, e.g. for tagged versions/special names. For example, how should the following invocation be resolved?
```
charlespierce: Agreed, I think the ambiguity makes supporting both methods a non-starter.

text/0000-update-version-handling.md
```diff
@@ -0,0 +1,225 @@
...
+
+```rust
+impl ToolSpec {
+  fn try_from_str(string: &str) -> Result<ToolSpec, String> {
+    // split apart the tool name and its version
+  }
+}
+```
+
+This is a relatively simple implementation. The only complication is the existence of Node namespaces: `@namespace/package`. This means parsing the version is not as simple as splitting on `"@"`. However, it should still be straightforward using the combination of a regular expression and the existing semver tooling.
```
charlespierce: I think we should also consult the code for `npm` and `yarn` to see how they do the splitting, so that we don't miss a corner case that we didn't even know existed. I _believe_, however, that the `@` character only represents a namespace if it's the first character in the name, anywhere else it can be treated as the delimiter between the version and the package name, so hopefully the logic shouldn't be too hard.

text/0000-update-version-handling.md
```diff
@@ -0,0 +1,225 @@
...
+
+The `ToolSpec` change will make the format `tool[@version]` work; updating the commands to take a `Vec<ToolSpec>` will support passing multiple arguments.
```
charlespierce: One important implementation detail (doesn't need to be added to the RFC itself, just so that we don't forget): We'll need to make sure when doing multiple `install` / `pin` operations that we do the `node` one (if present) _first_, since that sets up the environment.

text/0000-update-version-handling.md
charlespierce: Agreed. We'll also need to figure out how to handle tags on the node version, which doesn't have the same npm registry concepts that other tools have, since it's a level above `npm`.

text/0000-update-version-handling.md
```diff
@@ -0,0 +1,225 @@
...
+
+# Critique
+[critique]: #critique
+
+This introduces breaking changes—something which should not be done lightly! It does so under the assumption that fetching or especially installing or pinning multiple tools simultaneously is a relatively common operation. This assumption may be wrong. It may also be the case that implementing this—with the required changes to
```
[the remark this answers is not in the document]
chriskrycho: Oh, thank you! Cut off indeed! Will push a fix.

text/0000-update-version-handling.md
chriskrycho: Yep, I pushed an update with this in the table, but I missed that if mentioned it in the text. Will update. Thanks!

text/0000-update-version-handling.md
chriskrycho: I believe that’s correct and that’s precisely what I was planning. I’ll add a note indicating that intent.

text/0000-update-version-handling.md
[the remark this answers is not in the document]
chriskrycho: Fixed by a7e29c.

text/0000-update-version-handling.md
chriskrycho: @charlespierce interesting. I *think* we should still be able to support that, with precisely the same restrictions: `notion install "node@8 || 9"`. (I don't expect it really ever to come up, but maintaining support doesn't hurt us.)

text/0000-update-version-handling.md
chriskrycho: @charlespierce yeah, I think that warrants a further round of design. My inclination is actually to break apart the notion of versions there a bit: *package* versions are not identical with *platform* versions, though given how npm and Yarn work they overlap in complicated ways.

text/0000-update-version-handling.md
chriskrycho: Updated with this substantially fleshed out (d1ce7fa6); thanks for highlighting it!

text/0000-update-version-handling.md
chriskrycho: Moved this to a <b>Future Work</b> subsection (e183bf13).

text/0000-update-version-handling.md
chriskrycho: Great clarification; I've updated the language to indicate that this is a matter of *relative* frequency (959d3c6f)—users may not do `notion pin` very often, but when they do I think they're fairly likely to be pinning multiple tools (esp. when starting projects).

text/0000-update-version-handling.md
chriskrycho: This was a great callout, I implemented it [here](https://github.com/notion-cli/notion/pull/383/commits/e045d32ef174cf3cd1a102a8ade01f1dfdecacdb). Being able to just `#[derive(PartialEq, Eq, PartialOrd, Ord)]` on that enum and have it work automatically was ✨ magic ✨.

chriskrycho:
Implemented in notion-cli/notion#383.

text/0000-update-version-handling.md
```diff
@@ -0,0 +1,273 @@
...
+Users may regularly want to add multiple tools to their toolchain at the same time. They are also accustomed to being able to install multiple dependencies simultaneously with the `npm install` or `yarn add` commands, both in a package and globally.
+
+-   **Local installs:**
+
+    ```sh
+    npm install ember-cli@3.9.1 typescript@3
+    ```
+    
+    ```sh
+    yarn global ember-cli@3.9.1 typescript@3
```
benblank: NItpick: `yarn add`, not `yarn global`.

text/0000-update-version-handling.md
```diff
@@ -0,0 +1,273 @@
...
+
+Valid <i>versions</i> are any of the following formats:
+
+-   `node@10`
+-   `node@10.5`
+-   `node@10.5.2`
+
+Valid <i>version specs</i> are any valid SemVer spec, as already implemented in Notion. As a representative (not exhaustive) sample:
+
+-   `node@10.5*`
```
benblank: Nitpick: `10.5.*`, not `10.5*`.

text/0000-update-version-handling.md
```diff
@@ -0,0 +1,273 @@
...
+    Does it refer to the latest version of `create-react-app`, or the default version of `create-react-app` and [the `latest` package](https://www.npmjs.com/package/latest)? For all non-SemVer-string parses, this will be impossible to resolve.
+
+    The proposed design here is well-proven with other tools which install more than simply a single tool-and-version pair, e.g. `npm`, `yarn`, and `cargo-edit`, and it does not fall prey to these problems of indeterminacy.
+
+-   Leave things as they are. This has the upside of matching other Node version managers, all of which use a close variant on our current syntax. However, none of those tools support installing anything but Node, so the comparison is less telling than it might seem.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+-   How do we handle errors when installing multiple things?
```
benblank: I suggest we again match what npm and Yarn do, which appears to be to abort on the first error.

text/0000-update-version-handling.md
chriskrycho: Thank you; fixed!

rwjblue:
I think this has landed already (in https://github.com/volta-cli/volta/pull/383), and is included in (at least) volta@0.5.0.

Should we merge this RFC?

dherman:
Merging!

PR #33, opened 2019-05-02
Per-project configurations

charlespierce:
> Support Volta configuration of hooks on a per-project basis using a `.volta/hooks.json` file in the project.

[Rendered](https://github.com/charlespierce/rfcs/blob/per_project_config/text/0000-per-project-config.md)

### DIFF
--- a/text/0000-per-project-config.md
+++ b/text/0000-per-project-config.md
@@ -6,21 +6,21 @@
 # Summary
 [summary]: #summary
 
-Support Volta configuration of both hooks and platform information on a per-project basis using a `.voltarc.json` file in the project root.
+Support Volta configuration of hooks on a per-project basis using a `.volta/hooks.json` file in the project.
 
 # Motivation
 [motivation]: #motivation
 
 We currently only support Volta configuration on a per-user basis in the `hooks.toml` file. This file allows users to change the URLs that tools are downloaded from. However, since a user could be working on both an open-source project (that uses the public registry) and an internal project (that uses a private registry), we instead want to allow the users to specify these settings on a per-project basis.
 
-Additionally, we currently already support per-project settings for the platform (node, npm, and yarn versions), specified in the `toolchain` key inside of `package.json`. Multiple users have requested an alternate method in a separate file, for various reasons (see https://github.com/notion-cli/notion/issues/282 for some discussion of one use-case). As we are already looking at adding per-project hooks to a separate file, that file should also support specifying the platform information.
+To allow for the potential of additional per-project Volta configurations in the future, we want to store this file in a `.volta` directory at the project root.
 
 # Pedagogy
 [pedagogy]: #pedagogy
 
-We need to document the format of the `.voltarc.json` file, which will mostly match the features available in the existing `hooks.toml`, however it will also include the additional settings for `toolchain` similar to `package.json`. Additionally, we need to document and explain the precedence order for when multiple configs are specified (e.g. user-level _and_ project-level).
+We need to document the format of the `hooks.json` file, which will match the features available in the existing `hooks.toml`. We also need to document the existence of a `.volta` directory within a project for per-project configurations. Additionally, we need to document and explain the precedence order for when multiple configs are specified (e.g. user-level _and_ project-level).
 
-Additionally, for consistency of configuration, we should update the user-wide config stored in `~/.volta` from `hooks.toml` to also be named `.voltarc.json`, so that it's immediately obvious that it contains the same settings as the project-level configuration. This will be a breaking change that requires users to translate their settings from the old format to the new.
+Finally, for consistency of configuration, we should update the user-wide config stored in `~/.volta` from `hooks.toml` to also be named `hooks.json`, so that it's immediately obvious that it contains the same settings as the project-level configuration. This will not be a breaking change since the existing `hooks.toml` format wasn't documented and wasn't part of our public API.
 
 This feature will primarily be used for more advanced use-cases, as the default public repositories and `package.json` settings should cover the majority of open-source projects.
 
@@ -29,42 +29,27 @@ This feature will primarily be used for more advanced use-cases, as the default 
 
 ## User-level configuration file
 
-We will need to update the `HookConfig` to parse the user-level `.voltarc.json`, if it exists. We should also allow the user to specify the user platform (`node`, `npm`, and `yarn`) in that file. To that end, we can likely merge the existing `~/.volta/tools/user/platform.json` into `~/.volta/.voltarc.json` as our source of truth for the user-default selected versions. This may also allow us to simplify some of the logic around determining the currently active platform, as we can merge all of the available configs into a single result once, and then use that to execute a tool.
+We will need to update the `HookConfig` to parse the user-level `hooks.json`, if it exists, instead of the `hooks.toml` file.
 
 ## Configuration Precedence
 
-When configurations are specified in multiple places, we need to deep merge them in a specified order so that we get a single configuration to work from. The order will be defined generally as more specific configurations having precedence over more general configurations. This means:
-
-- Per-project configuration beats per-user configuration
-- `.voltarc.json` platform spec must match `package.json` platform spec if both are specified
+When configurations are specified in multiple places, we need to deep merge them in a specified order so that we get a single configuration to work from. The order will be defined generally as more specific configurations having precedence over more general configurations. This means that per-project settings will beat per-user settings.
 
 ## Merging Configurations
 
 When multiple configurations are specified, we should merge them deeply using the above precedence such that the final configuration includes all hooks defined by any of the configurations. For example, if:
 
-- User-level `.voltarc.json` includes the `[node.index]` and `[node.distro]` hooks
-- Project-level `.voltarc.json` includes the `[node.distro]` hook
+- User-level `hooks.json` includes the `[node.index]` and `[node.distro]` hooks
+- Project-level `.volta/hooks.json` includes the `[node.distro]` hook
 
 The final configuration that Volta uses will have:
 
-- `[node.index]` hook from user-level `.voltarc.json`
-- `[node.distro]` hook from project-level `.voltarc.json`
-
-This will likely require that the `Project` and `Hooks` objects in `volta-core` are merged in some way into a single configuration object that pulls from all possible sources.
-
-Additionally, if the platform spec is provided in both `.voltarc.json` _and_ in `package.json`, they must be the same or we will throw an error. This will prevent confusion around which file "wins" and what the source of truth is for the project. The documentation should strongly recommend that one or the other is specified, not both.
+- `[node.index]` hook from user-level `hooks.json`
+- `[node.distro]` hook from project-level `hooks.json`
 
 ## `volta config` Reference
 
 To help users understand which configurations are active and where they are coming from, we should implement a `volta config list` command that works similarly to `npm config list`. This will allow users to immediately see what Volta understands about the state of the world from a given place in the filesystem, while also providing us with a useful tool for debugging the implementation. This also opens the option going forward of providing CLI commands for editing the configuration (e.g. `volta config set`).
-
-## Multi-stage Implementation
-
-To facilitate the work and allow users to work with parts of these changes before we make an entire overhaul to the system, we can implement this RFC in multiple stages:
-
-1. Add support for parsing a project-local `volta-hooks.toml` file, if it exists, as well as the `~/.volta/hooks.toml` that we currently check for.
-2. Rename `hooks.toml` into `.voltarc.json`, changing the file format but supporting the same set of configurations (i.e. Only hooks, no platform specifications at this point).
-3. Fully merge the hooks config with the platform config internally and add support for `toolchain` within `.voltarc.json`
 
 # Critique
 [critique]: #critique
@@ -73,15 +58,11 @@ To facilitate the work and allow users to work with parts of these changes befor
 
 One possible update is that a different precedence order could be used for the configurations. However, it is a common pattern in computing that more specific settings beat less specific settings, so we should still adhere to that. Additionally, the specific-to-general precedence order matches the order used by `npm` for `npmrc` files.
 
-As discussed above, the project-level `.voltarc.json` setting is considered to be a more "advanced" way of specifying per-project settings, so it should have precedence over `package.json`.
-
 ## Merging Configurations
 
-Another approach to determining the final configuration from the available files would be to use _only_ the highest-precedence `.voltarc.json` if it is present and fall back to the existing `package.json` if there isn't a `.voltarc.json` present. This would mean that if any `.voltarc.json` is present, it represents the single source of truth for all of the Volta settings. However, it would remove the ability to have settings cascade down from the user-level settings to the project-level settings when a project doesn't have any custom configuration beyond what the user has specified.
+Another approach to determining the final configuration from the available files would be to use _only_ the highest-precedence `hooks.json` if it is present, This would mean that if any `hooks.json` is present, it represents the single source of truth for all of the Volta hooks. However, it would remove the ability to have settings cascade down from the user-level settings to the project-level settings when a project doesn't have any custom configuration beyond what the user has specified.
 
 The merge approach suggested in the RFC matches how `npm` resolves configuration from multiple sources, pulling each value in precedence order, but allowing different files to specify different settings.
 
 # Unresolved questions
 [unresolved]: #unresolved-questions
-
-- What additional settings, if any, should we include in `.voltarc.json`?
--- a/text/0000-per-project-config.md
+++ b/text/0000-per-project-config.md
@@ -64,6 +64,10 @@ Another approach to determining the final configuration from the available files
 
 The merge approach suggested in the RFC matches how `npm` resolves configuration from multiple sources, pulling each value in precedence order, but allowing different files to specify different settings.
 
+## File Layout
+
+An alternative to using a `.volta` directory with independent files would be to have a single `.voltarc` or `.voltarc.json` file in the root of the project. The trouble with this approach is that it would require any future project-level configuration we decide to add to also be in the same file. In a corporate setting, it's reasonable to expect that different groups will be responsible for managing different configurations (i.e. one group is in charge of changes to the hooks while another group manages the `node` platform settings). In this case, it would be preferable to have the files be separate, so to future-proof the design, we decided on having a configuration directory that can hold any number of independent files.
+
 ## File Format
 
 Instead of using `.json`, we could use a format that supports comments for our configuration files, such as `.toml`. This would have the advantage of allowing the configuration to be more expressive, but would require a learning curve for a majority of our users, who don't necessarily have experience with `.toml`. By using `.json`, we're using a format that is a standard in the JS ecosystem, so the learning curve will be lowered. Additionally, there are proposals for extensions of JSON that support comments, so in the future we could adopt those in a backwards-compatible way, which would allow us to support comments without breaking existing users.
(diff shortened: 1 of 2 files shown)


text/0000-per-project-config.md
```diff
@@ -0,0 +1,67 @@
...
+
+Additionally, we currently already support per-project settings for the platform (node and yarn versions), specified in the `toolchain` key inside of `package.json`. Multiple users have requested an alternate method in a separate file, for various reasons (see https://github.com/notion-cli/notion/issues/282 for some discussion of one use-case). As we are already looking at adding per-project hooks to a separate file, that file should also support specifying the platform information.
+
+# Pedagogy
+[pedagogy]: #pedagogy
+
+We need to document the format of the `.notionrc` file, which will mostly match the format of `hooks.toml`, however it will also include the additional settings for `node` and `yarn` versions. Additionally, we need to document and explain the precedence order for when multiple configs are specified (e.g. user-level `hooks.toml` _and_ project-level `.notionrc`).
```
benblank: > settings for `node` and `yarn` versions

We also support `npm` in `package.json`, if I'm not mistaken.

text/0000-per-project-config.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Merging Configurations
+
+Another approach to determining the final configuration from the available files would be to use _only_ `.notionrc` if it is present and fall back to the existing combination of `package.json` and `hooks.toml` if there isn't a `.notionrc` present. This would mean that if `.notionrc` is present, it represents the single source of truth for all of the Notion settings. However, it would remove the ability to have settings cascade down from the user-level settings to the project-level settings when a project doesn't have any custom configuration beyond what the user has specified.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- Should we also include additional settings in `.notionrc`?
+- Is there a better file name to use, other than `.notionrc`? `notion.toml` Perhaps?
```
benblank: It's becoming increasingly common in JS-land (for example, [.eslintrc.yaml](https://eslint.org/docs/user-guide/configuring) and [.babelrc.js](https://babeljs.io/docs/en/config-files)) to include the extension in the rc name, e.g. `.notionrc.toml`.  It plays nicer with editors' file-type detection.

text/0000-per-project-config.md
```diff
@@ -0,0 +1,67 @@
...
+- Is TOML the best format, or should we use something else for this per-project configuration?
```
benblank: My preference would be JSON, as having more common editor support and better user familiarity among JS developers.

text/0000-per-project-config.md
thoov: > multiple configs are specified (e.g. user-level `hooks.toml` _and_ project-level 
>`.notionrc`).

We should make sure that all configs have the same name

text/0000-per-project-config.md
thoov: I second either JSON or YAML as those would be the most familiar with JS devs.

text/0000-per-project-config.md
thoov: I agree with this

text/0000-per-project-config.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Configuration Precedence
+
+One possible update is that a different precedence order could be used for the configurations. However, it is a common pattern in computing that more specific settings beat less specific settings, so we should still adhere to that. As discussed above, the `.notionrc` setting is considered to be a more "advanced" way of specifying per-project settings, so it should have precedence over `package.json`.
+
+## Merging Configurations
+
+Another approach to determining the final configuration from the available files would be to use _only_ `.notionrc` if it is present and fall back to the existing combination of `package.json` and `hooks.toml` if there isn't a `.notionrc` present. This would mean that if `.notionrc` is present, it represents the single source of truth for all of the Notion settings. However, it would remove the ability to have settings cascade down from the user-level settings to the project-level settings when a project doesn't have any custom configuration beyond what the user has specified.
+
+# Unresolved questions
```
thoov: We should add a command to notion: `notion config list` that acts like yarn or npm does where it shows the merged config at that location. This would be helpful to debug / verify things.

text/0000-per-project-config.md
```diff
@@ -0,0 +1,67 @@
...
+- `[node.distro]` hook and `node` version from `.notionrc`
+
+This will likely require that the `Project` and `Hooks` objects in `notion-core` are merged in some way into a single configuration object that pulls from all possible sources.
+
+# Critique
+[critique]: #critique
+
+## Configuration Precedence
+
+One possible update is that a different precedence order could be used for the configurations. However, it is a common pattern in computing that more specific settings beat less specific settings, so we should still adhere to that. As discussed above, the `.notionrc` setting is considered to be a more "advanced" way of specifying per-project settings, so it should have precedence over `package.json`.
```
thoov: We should try and be as close to how npm/yarn does it: https://docs.npmjs.com/files/npmrc#files

text/0000-per-project-config.md
thoov: I agree with this

text/0000-per-project-config.md
thoov: I second either JSON or YAML as those would be most familiar with JS devs

text/0000-per-project-config.md
charlespierce: We nominally support it, but it doesn't actually work. `npm` currently has to be set to the version that ships with `node`, and if it isn't then things will break. That said, you're right that we should include it in the discussion because the plan is for it to work eventually.

text/0000-per-project-config.md
charlespierce: Okay, it looks like that matches with the proposed precedence order and merge technique in this RFC. I'll update to reflect that the suggested behavior matches what `npm` does

text/0000-per-project-config.md
charlespierce: Good suggestion! We had a `notion config` command for a long time, but it never actually did anything. I'll add a section to the details about it, though we might need another RFC specifically for that command.

text/0000-per-project-config.md
charlespierce: I like that, then you get the 'hidden' aspect of dotfiles but still have an extension to make it immediately obvious what format should be used.

text/0000-per-project-config.md
charlespierce: We decided to stick with TOML for the hooks configuration in https://github.com/notion-cli/rfcs/pull/25, however since we are now proposing to support the platform settings from `package.json`, I think using JSON makes the most sense.

text/0000-per-project-config.md
charlespierce: Agreed 👍

text/0000-per-project-config.md
```diff
@@ -0,0 +1,79 @@
...
+
+## User-level configuration file
+
+We will need to update the `HookConfig` to parse the user-level `.notionrc.json`, if it exists. We should also allow the user to specify the user platform (`node`, `npm`, and `yarn`) in that file. To that end, we can likely merge the existing `~/.notion/tools/user/platform.json` into `~/.notion/.notionrc.json` as our source of truth for the user-default selected versions. This may also allow us to simplify some of the logic around determining the currently active platform, as we can merge all of the available configs into a single result once, and then use that to execute a tool.
+
+## Configuration Precedence
+
+When configurations are specified in multiple places, we need to merge them in a specified order so that we get a single configuration to work from. The order will be defined generally as more specific configurations having precedence over more general configurations. This means:
+
+- Per-project configuration beats per-user configuration
```
stefanpenner: is this a deep or shallow merge

charlespierce: My mental model was a deep merge (which I think shows in the example later on), I'll call that out explicitly, however.

text/0000-per-project-config.md
```diff
@@ -0,0 +1,87 @@
...
+## User-level configuration file
+
+We will need to update the `HookConfig` to parse the user-level `.notionrc.json`, if it exists. We should also allow the user to specify the user platform (`node`, `npm`, and `yarn`) in that file. To that end, we can likely merge the existing `~/.notion/tools/user/platform.json` into `~/.notion/.notionrc.json` as our source of truth for the user-default selected versions. This may also allow us to simplify some of the logic around determining the currently active platform, as we can merge all of the available configs into a single result once, and then use that to execute a tool.
+
+## Configuration Precedence
+
+When configurations are specified in multiple places, we need to deep merge them in a specified order so that we get a single configuration to work from. The order will be defined generally as more specific configurations having precedence over more general configurations. This means:
+
+- Per-project configuration beats per-user configuration
+- `.notionrc.json` platform spec beats `package.json` platform spec
```
mikrostew: Do we still need the `"toolchain"` in `package.json` if there can be the same info in the same format in `.notionrc.json`? It seems too complicated to have two possible configurations in the same project, with differing precedence.

I'm thinking that part of the multi-stage implementation could be to deprecate the `package.json` config and move everything into `.notionrc.json`, to have one source of truth in the project.

text/0000-per-project-config.md
```diff
@@ -0,0 +1,87 @@
...
+## Merging Configurations
+
+Another approach to determining the final configuration from the available files would be to use _only_ the highest-precedence `.notionrc.json` if it is present and fall back to the existing `package.json` if there isn't a `.notionrc.json` present. This would mean that if any `.notionrc.json` is present, it represents the single source of truth for all of the Notion settings. However, it would remove the ability to have settings cascade down from the user-level settings to the project-level settings when a project doesn't have any custom configuration beyond what the user has specified.
+
+The merge approach suggested in the RFC matches how `npm` resolves configuration from multiple sources, pulling each value in precedence order, but allowing different files to specify different settings.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- What additional settings, if any, should we include in `.notionrc.json`?
```
mikrostew: I think this is out of scope for this RFC, but in the future we might want to add config for a version specifier for `notion`? In case the project uses some Notion feature that is only available after a certain release?

text/0000-per-project-config.md
```diff
@@ -0,0 +1,87 @@
...
+
+Additionally, we currently already support per-project settings for the platform (node, npm, and yarn versions), specified in the `toolchain` key inside of `package.json`. Multiple users have requested an alternate method in a separate file, for various reasons (see https://github.com/notion-cli/notion/issues/282 for some discussion of one use-case). As we are already looking at adding per-project hooks to a separate file, that file should also support specifying the platform information.
+
+# Pedagogy
+[pedagogy]: #pedagogy
+
+We need to document the format of the `.notionrc.json` file, which will mostly match the features available in the existing `hooks.toml`, however it will also include the additional settings for `toolchain` similar to . Additionally, we need to document and explain the precedence order for when multiple configs are specified (e.g. user-level _and_ project-level).
```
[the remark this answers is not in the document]
charlespierce: Good catch, thanks!

text/0000-per-project-config.md
charlespierce: I think there is definitely still value in supporting the `toolchain` key in `package.json`. Specifically in the (likely very common) case where you don't need to specify any custom hooks, then you don't have to add a file to your project.

It does introduce some complexity in technically allowing you to have different configurations in the same project, but given that `.notionrc.json` will always be a more advanced configuration, I think the benefits of easier onboarding in the simple case outweight the negatives of the more complicated advanced case. We could also mitigate that by including some warnings when performing the merge that the user has two different values specified and that they should remove the `package.json` one since it isn't being used anyway.

text/0000-per-project-config.md
charlespierce: Definitely, that's a great idea and will allow a nice, actionable error message early in the process, instead of some opaque one later when something fails. Or even worse, doing the wrong thing because we don't even read a new configuration at all.

text/0000-per-project-config.md
thoov: I personally am a fan of consolidating things into `.notionrc.json`. Though if we keep `toolchain`, notion should just error out if the values are different than the ones specified in `.notionrc.json`. That seems like the easiest to implement and doesn't put the users into a weird state.

charlespierce:
@rwjblue I think the performance penalties are definitely something to keep in mind, but I'm not sure a `notion run` command covers the case of wanting a single source-of-truth for the node version that can be used for build caching as described in that issue.

Additionally, while it's definitely true that parsing extra files has a cost, in this case at least, the cost is limited to one extra parse, and not N like in the case of parsing the dependent `package.json` files. We also already have to do the discovery of the `package.json`, and we can require that the new file be a sibling of that so as not to have to double-up the speculative file-structure walking.

Also worth noting: Because we have event reporting hooks in `hooks.toml`, we actually _do_ parse that file on every invocation (though for the general case it's not parsed until _after_ the command has run), so in practice there won't be a difference between the two (though that can certainly change if we decide the performance hit is too great).

text/0000-per-project-config.md
```diff
@@ -0,0 +1,87 @@
...
+## Merging Configurations
+
+Another approach to determining the final configuration from the available files would be to use _only_ the highest-precedence `.voltarc.json` if it is present and fall back to the existing `package.json` if there isn't a `.voltarc.json` present. This would mean that if any `.voltarc.json` is present, it represents the single source of truth for all of the Volta settings. However, it would remove the ability to have settings cascade down from the user-level settings to the project-level settings when a project doesn't have any custom configuration beyond what the user has specified.
+
+The merge approach suggested in the RFC matches how `npm` resolves configuration from multiple sources, pulling each value in precedence order, but allowing different files to specify different settings.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- What additional settings, if any, should we include in `.voltarc.json`?
```
chriskrycho: I'd like to suggest that using a comment-able file format (e.g. `TOML`) should at least be put in <b>Unresolved Questions</b> or alternatives. While there's a large advantage to using the same format everybody in the ecosystem uses, the fact that you can't put comments explaining anything in your `.json` files is a serious pain point there. There would be a small but real teaching burden for `.volta.toml` but I think it's at least worth noting as an alternative and *explicitly* deciding against it.

rwjblue:
> Because we have event reporting hooks in hooks.toml, we actually do parse that file on every invocation

Sorry, why do we need hooks.toml for every invocation of a tool (either `notion install`'ed or `node`/`yarn`)? I'm not intimately familiar with what we store there  that would be needed...

dherman:
@rwjblue For event reporting, e.g. when you've hooked up Volta to log diagnostic information for internal reporting/monitoring.

dherman:
Let me map out some of the constraints first, and then I'll conclude with a proposal.

# Constraints

## Category: Syntax

**Constraint:** *Refactoring*
We should try to stick to consistent configuration file syntax in all places that the same kinds of configuration information might live (e.g. project-local vs user) so it's as easy as possible to refactor.

**Constraint:** *Learnability*
The config syntax should be as readily understandable as possible for JS developers.

**Constraint:** *Comments*
We should support a way to write comments in configuration files. (Why, Crock, why?!)


## Category: Filesystem layout

**Constraint:** *Separate layers and ownership*
Tool pinning and site-management hooks represent different roles, which will often be performed by different people. The former is managed by the development team when choosing or upgrading tools. The latter is managed by the team responsible for policies around where tools are provisioned from, which is often an IT or infrastructure org.

For learnability, readability, and independent maintenance, it’s best to keep configuration relating to hooks in separate files from configuration relating to pinned tools.

**Constraint:** *Low adoption overhead*
For smaller open source repos, a mandatory extra top-level file or directory would impose cognitive overhead and be resented by people considering adopting Volta.

**Constraint:** *Future-proofing for more configs*
We should future-proof against the possibility of *ever* adding multiple Volta files to the top-level, even if they're opt-in.

**Constraint:** *Fast path performance*
Complex filesystem lookups could impose latency costs on fast paths. The design should not conflict with Volta's fast-path performance goals.

**Constraint:** *Single source of truth*
If we allow multiple possible representations for identical information (e.g. `package.json` and a project-local config file), we should at least not allow inconsistent data to coexist in more than one location.

**Constraint:** *Workspace support*
It should be possible to support workspaces (aka monorepos) by mirroring the Yarn workspace algorithm, allowing configuration and tool pinning to be shared across multiple projects within a workspace.

**Constraint:** *Ready-made cache key for Docker*
It'd be nice to have a standardized file in source control that contains your basic tool versions and nothing else, so that the file can be used as a cache key for layer caching in a `Dockerfile`. This allows Docker users to just `COPY` the file from source control directly in their `Dockerfile` and then install the tools that correspond to it, and the cache will never be invalidated as long as those tool choices don't change.

Otherwise, the way to provide that data as inputs is by extracting the information with shell commands and passing it along either through generated files, environment variables, or Docker build arguments. But not every system that provides support for `Dockerfile` has the right hooks to allow performing the shell commands at the right point, which then ends up requiring users to do manual steps and keep up to date with the source of truth.

# Proposal

Putting these considerations together brings me to:

## Syntax (now)

The RFC I think correctly identifies that JSON is the most familiar config syntax to use. This doesn’t allow comments, but see the next section. I think for now, the amount of config is small enough that lack of comments isn’t a high priority issue.

## Syntax (future)

I think we can just add support for comments. JSON + comments is an [increasingly common phenomenon](https://github.com/Microsoft/TypeScript/issues/4987), and how much I care what Crock thinks about it + $4.15 will get you a Venti Latte at Starbuck’s. 😛 

There are already [Serde implementations](https://github.com/hjson/hjson-rust) of syntaxes like [HJSON](http://hjson.org). This would be a backwards-compatible change so we don’t need to block on figuring out exactly which superset of JSON we want to support.


## Filesystem layout (now)

Tool pinning should only be supported in `package.json` for now, maintaining the separation of developer-facing tool pinning information from low-level site configuration.

This doesn’t solve the Docker problem but see the next section. I think for now, there are workarounds people can use in those setups (e.g. automatically generating the cache key file they want as part of their build step, as a git trigger, etc).

Other per-project configuration should go in a `.volta` directory at project top-level, future-proofing for additional configuration files down the road. For now, the hooks should go in `.volta/hooks.json`. If we want to commit to a particular human-JSON dialect now, we can choose a different filename extension, but the `tsconfig.json` precedent suggests to me that it’s OK to stick with `.json`.


## Possible extensions (future)

In the future, we could support the Docker use case by allowing projects to opt in to using a tool lockfile instead of `package.json`. This would live in `.volta/tools.lock`. We would want to test for latency regressions, but I suspect it can be made negligible.

I could imagine simple commands like `volta pin --lockfile` as a way of opting in.

I believe that workspace support should be something we can add without significant incompatibility—basically, the resolution algorithm would search further up for workspace roots, to allow more sharing. So it’s not an immediately blocking issue, but it’s reasonably high priority since workspaces are popular.

chriskrycho:
One initial **question-for-clarification** on that proposal:

> Tool pinning should only be supported in `package.json` for now, maintaining the separation of developer-facing tool pinning information from low-level site configuration.

Here, I'm wondering what you mean by "for now"—i.e. for the next little bit while we have other higher priorities? Or indefinitely unless/until we do a further RFC in that direction? My own suggestion would be to *include* it as a planned extension of the current behavior, but with a known lower prioritization. In that case, we would need a small bit of additional design to specify which "wins" if both are specified.

One **thought**, related to that:

> We would want to test for latency regressions, but I suspect it can be made negligible.

In principle I think these could "just" be run in parallel and so we'd only be waiting on the I/O and parse of the slower of them – almost certainly `package.json`. (I'm curious if anyone sees a reason we couldn't do that.)

dherman:
@chriskrycho 

> My own suggestion would be to include it as a planned extension of the current behavior, but with a known lower prioritization.

I'm not sure I have a strong inclination either way. My only thought is that I'm not confident we've really surfaced all the constraints around why people would want a separate lockfile. Maybe the Docker layer caching use case is enough? I just wonder if it might want some time for more users to onboard and surface use cases and constraints.

text/0000-per-project-config.md
charlespierce: Good call, added!

charlespierce:
Updated with limited scope (only hooks) and with an additional critique section about the file format.

PR #34, opened 2019-05-15
RFC: introduce `volta list` and deprecate `volta current`

chriskrycho:
[Rendered RFC text](https://github.com/volta-cli/rfcs/blob/info-commands/text/0000-informational-commands.md). See earlier discussion at volta-cli/volta#349.

> Add an informational command, `volta list`, replacing the `volta current` command and expanding on its capabilities. The `list` command will allow users to see their default and project-specific toolchains, and will support both human-friendly output and a tool-friendly output.

### DIFF
--- a/text/0000-informational-commands.md
+++ b/text/0000-informational-commands.md
@@ -511,9 +511,12 @@ This section discusses the tradeoffs considered in this proposal. This must incl
 [unresolved]: #unresolved-questions
 
 - Should the bare `list` command include:
+
     - all available package binaries?
     - a subset of package binaries once the number crosses a threshold, with instructions about how to see all?
     - no package binaries?
+
+    And how should that be presented in `plain` vs. `human` mode?
 
 <!-- 
 - What parts of the design do you expect to resolve through the RFC process before this gets merged?
--- a/text/0000-informational-commands.md
+++ b/text/0000-informational-commands.md
@@ -758,7 +758,7 @@ We could leave the current state of affairs as it is. However, users then have n
 # Unresolved questions
 [unresolved]: #unresolved-questions
 
-- Should the bare `list` command include:
+- Should the bare command include:
 
     - all available package binaries?
     - a subset of package binaries once the number crosses a threshold, with instructions about how to see all?
@@ -768,11 +768,7 @@ We could leave the current state of affairs as it is. However, users then have n
 
 - How should the current version be identified? It is currently marked with `(current @ <path>)`. Should this be `(from <path>)` or some other design?
 
-- Should the `list` command accept flags to narrow the search instead of subcommands, e.g. `volta list --node`, `volta list --yarn`, `volta list --package=typescript`, etc.?
-
-    - Is there value in supporting both the `tool` and `package` invocations, or should we collapse them and avoid the additional complexity of distinguishing between tools/binaries and their supplying packages?
-
-    - Should we simply parse a positional argument after `volta list` as `<node|npm|yarn|<package name>>`?
+- Should the command accept flags to narrow the search instead of a positional argument, e.g. `volta list --node`, `volta list --yarn`, `volta list --package=typescript`, etc.?
 
 - How should the built-in version of `npm` be displayed to the user? Options include:
     - associated with the runtime it ships with, e.g. `v12.2.0 (with npm v6.4.1)`
--- a/text/0000-informational-commands.md
+++ b/text/0000-informational-commands.md
@@ -719,6 +719,10 @@ This section discusses the tradeoffs considered in this proposal. This must incl
 
 - Should the `list` command accept flags to narrow the search instead of subcommands, e.g. `volta list --node`, `volta list --yarn`, `volta list --package=typescript`, etc.?
 
+- How should the built-in version of `npm` be displayed to the user? Options include:
+    - associated with the runtime it ships with, e.g. `v12.2.0 (with npm v6.4.1)`
+    - as a packager, but indicating its source as a built-in, e.g. `npm v6.4.1 (built-in from node@v12.2.0`
+
 <!-- 
 - What parts of the design do you expect to resolve through the RFC process before this gets merged?
 - What parts of the design do you expect to resolve through the implementation of this feature before stabilization?
--- a/text/0000-informational-commands.md
+++ b/text/0000-informational-commands.md
@@ -42,7 +42,7 @@ Throughout, we will assume the user has the following configuration for illustra
 
 They also have two projects with the following pins:
 
-- `~/pinned/node-only/package.json`:
+- `~/node-only/package.json`:
 
     ```json
     {
@@ -52,7 +52,7 @@ They also have two projects with the following pins:
     }
     ```
 
-- `~/pinned/node-and-yarn/package.json`:
+- `~/node-and-yarn/package.json`:
 
     ```json
     {
@@ -75,8 +75,8 @@ The format is:
 $ volta list
 ⚡️ Currently active toolchain:
 
-    Node: version [(default|project)]
-    <packager>: <built-in|version> [(default|project)]]
+    Node: version [(default|<project path>)]
+    <packager>: <built-in|version> [(default|<project path>)]]
     Tool binaries available:
         <comma separated list|NONE>
 
@@ -107,7 +107,7 @@ See options for more detailed reports by running `volta list --help`.
 $ volta list
 ⚡️ Currently active toolchain:
 
-    Node: v8.16.0 (from `~/pinned/node-only/package.json`)
+    Node: v8.16.0 (from `~/node-only/package.json`)
     Yarn: v1.12.3 (default)
     Tool binaries available:
         create-react-app, ember, tsc, tsserver, yarn-deduplicate
@@ -123,8 +123,8 @@ See options for more detailed reports by running `volta list --help`.
 $ volta list
 ⚡️ Currently active toolchain:
 
-    Node: v12.2.0 (from `~/pinned/node-and-yarn/package.json`)
-    Yarn: v1.16.0 (from `~/pinned/node-and-yarn/package.json`)
+    Node: v12.2.0 (from `~/node-and-yarn/package.json`)
+    Yarn: v1.16.0 (from `~/node-and-yarn/package.json`)
     Tool binaries available:
         create-react-app, ember, tsc, tsserver, yarn-deduplicate
 
@@ -142,15 +142,15 @@ The basic format is:
 ```sh
 $ volta list --all
 Node runtimes:
-    <version> [(default|project)]
+    <version> [(default|<project path>)]
 
 Packagers:
     (Yarn|npm):
-        <version> [(default|project)]
+        <version> [(default|<project path>)]
 
 Tools:
     <tool name>
-        <version> [(default|project)]
+        <version> [(default|<project path>)]
             binaries: [<binary name>]...
             platform:
                 runtime: node@<version>
@@ -270,7 +270,7 @@ Tools:
 ```sh
 $ volta list --all
 Node runtimes:
-    v12.2.0 (project)
+    v12.2.0 (`~/node-and-yarn/project.json`)
     v11.9.0
     v10.15.3 (default)
     v8.16.0
@@ -278,7 +278,7 @@ Node runtimes:
 Packagers:
     Yarn:
         v1.16.0 (default)
-        v1.12.3 (project)
+        v1.12.3 (`~/node-and-yarn/project.json`)
 
 Tools:
     create-react-app:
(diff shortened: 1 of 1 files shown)


text/0000-informational-commands.md
```diff
@@ -0,0 +1,302 @@
...
+
+Throughout, we will assume the user has the following configuration for illustrative purposes:
+
+- Node versions installed: v12.2.0, v11.9.0, v10.15.3 (default), v8.16.0
+- Yarn versions installed: v1.16.0, v1.12.3 (default)
+- Tools installed:
+    - ember-cli: v3.10.0 (default) on Node v12.2.0, v3.8.2 on Node v12.2.0, with binary `ember`
+    - typescript: v3.4.5 on Node v12.2.0, v3.0.3 (default) on Node v12.2.0, with binaries `tsc`, `tsserver`
+    - create-react-app: v3.0.1 (default) on Node v12.2.0, with binary `create-react-app`
+    - yarn-deduplicate: v1.1.1 on Node v12.2.0, with binary `yarn-deduplicate` (notice that this is not a *default*; assume the user ran `volta )
```
rwjblue:
```suggestion
    - yarn-deduplicate: v1.1.1 on Node v12.2.0, with binary `yarn-deduplicate` (notice that this is not a *default*; assume the user ran `volta fetch` )
```

text/0000-informational-commands.md
```diff
@@ -0,0 +1,302 @@
...
+## Overview of CLI commands
+
+Top-level command: show the current to
+
+```sh
+$ volta list
+⚡️ Currently active toolchain:
+
+    Node: v8.16.0 (from `~/path/to/project/package.json`)
+    Yarn: v1.12.3 (default)
```
rwjblue: Note: https://github.com/volta-cli/volta/issues/436, this is not actually possible (to use a project pinned node but non-project pinned yarn).

text/0000-informational-commands.md
```diff
@@ -0,0 +1,302 @@
...
+
+```sh
+$ volta list --all
+Node runtimes:
+    v12.2.0
+    v11.9.0
+    v10.15.3 (default)
+    v8.16.0
+
+Packagers:
```
rwjblue: This list should probably include `npm` as well, since every node comes with one.

text/0000-informational-commands.md
chriskrycho: Interesting; from a conversation with @charlespierce the other day I thought it *was* supported; I'll tweak this accordingly. Good catch; and thank you!

text/0000-informational-commands.md
chriskrycho: That's an interesting point I've been mulling on as well. My thinking here was that we'd be listing *custom* packagers, rather than the defaults – which probably means listing it with the runtimes? But I'm very much up for changing my mind on how it should be handled.

text/0000-informational-commands.md
```diff
@@ -0,0 +1,525 @@
...
+runtime node@<version> (default|<project path>)
+packager <npm|yarn>@<version> (built-in|default|<project path>)
+```
+
+<details><summary>Outside a project</summary>
+
+```sh
+$ volta list
+runtime node@v10.15.3 (default)
+packager yarn@v1.12.3 (default)
```
chriskrycho: I'm unsure what we should actually output here: if the user has `install`-ed Yarn, presumably we should print it. Should we also always print the built-in `npm`? I currently lean toward yes; that would then extend to the `human` format as well. Thoughts?

text/0000-informational-commands.md
chriskrycho: Added an unresolved question on this point.

text/0000-informational-commands.md
```diff
@@ -0,0 +1,775 @@
...
+        binaries: [<binary name>]...
+        platform:
+            runtime: node@<version>
+            packager: built-in npm|<npm|yarn>@<version>
+```
+
+For the TypeScript config specified in the canonical example:
+
+```sh
+volta list package typescript --human
```
mikrostew: Slight discrepancy here between `volta list <package>` and `volta list package typescript`.

This will probably need to be `volta list package <package>` and `volta list tool <tool>` to be able to disambiguate when packages and tools have the same names (like the `create-react-app` example).

text/0000-informational-commands.md
```diff
@@ -0,0 +1,775 @@
...
+We could leave the current state of affairs as it is. However, users then have no insight into their toolchain: they must inspect the internals of `~/.volta`  to see what they actually have on their system, and in order to see what version of a tool is going to be used, they have must actually invoke it or run `volta which` and interpret the path.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- Should the bare command include:
+
+    - all available package binaries?
+    - a subset of package binaries once the number crosses a threshold, with instructions about how to see all?
+    - no package binaries?
```
mikrostew: Currently the only way to create a shim for a binary is to use `volta install <package>`, which installs the package and shims any of the installed binaries. Then the shim will intercept any invocation inside of a project to detect if it is a package binary.

Since this is already displaying the available binaries (from installed packages), the available package binaries will be a subset of that, and could probably be displayed with some indication that they are coming from the current project.

text/0000-informational-commands.md
```diff
@@ -0,0 +1,775 @@
...
+
+    And how should that be presented in `plain` vs. `human` mode? Should there be any differences between `plain` and `human` for the bare command (as in the current design)?
```
mikrostew: My first thought at this would be the addition of a single line (if for instance the project uses ember):

```
$ volta list --human
⚡️ Currently active tools:
    Node: v8.16.0 (current @ ~/node-only/package.json)
    Yarn: v1.12.3 (default)
    Tool binaries available:
        create-react-app, tsc, tsserver
        ember (current @ ~/node-only/package.json)
```

text/0000-informational-commands.md
```diff
@@ -0,0 +1,775 @@
...
+    ```
+
+### `volta list` (no flags)
+
+#### Human
+
+The format is:
+
+```sh
+$ volta list --human
```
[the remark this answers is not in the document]
chriskrycho: Ah, yes, this was a case where I forgot to update the rest of the text after realizing that it's much easier to do `--print=<more modes>`, and concluding it's probably better—because I don't think it's something you'll type often at all. I'll update the rest of the RFC text accordingly.

text/0000-informational-commands.md
```diff
@@ -0,0 +1,775 @@
...
+        "yarn": "v1.16.0"
+      }
+    }
+    ```
+
+### `volta list` (no flags)
+
+#### Human
+
+The format is:
```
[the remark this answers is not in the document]
chriskrycho: Ah, good catch – thank you; I'll fix it tomorrow!

text/0000-informational-commands.md
```diff
@@ -0,0 +1,775 @@
...
+- Yarn versions installed: v1.16.0, v1.12.3 (default)
+- Tools installed:
+    - ember-cli, with binary `ember`:
+	    - v3.10.0 on Node v12.2.0 with built-in npm (default)
+	    - v3.8.2 on Node v12.2.0 with built-in npm
+    - typescript, with binaries `tsc`, `tsserver`:
+	    - v3.4.5 on Node v12.2.0 with built-in npm
+	    - v3.0.3 on Node v12.2.0 with built-in npm (default)
+    - create-react-app, with binary `create-react-app`: v3.0.1 on Node v12.2.0 with built-in npm (default)
+    - yarn-deduplicate, with binary `yarn-deduplicate`: v1.1.1 on Node v12.2.0 with built-in npm (notice that this is not a *default*; assume the user ran `volta fetch` )
```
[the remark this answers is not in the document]
chriskrycho: Ah, that's great to know – I'd missed that detail. I'll update!

text/0000-informational-commands.md
chriskrycho: This is actually the reason why I had `volta list package <package>` and `volta list tool <tool>` in an earlier draft of the RFC. I'm back-and-forth on it, because in the case you identified, it's the same output either way. In a case where there's more than one binary, but one does match the package name, I don't think it (harms* anything to default to the package in that case? 🤔 But I'm *very* willing to be persuaded here—I just found myself leaning *slightly* in the direction of brevity!

text/0000-informational-commands.md
chriskrycho: I like that design a lot. Should the previous line then include `(defaults)` to distinguish them? 🤔

text/0000-informational-commands.md
chriskrycho: This plus your suggestion below make a ton of sense to me; I'll update the design accordingly and at least the initial implementation will work that way.

text/0000-informational-commands.md
```diff
@@ -0,0 +1,775 @@
...
+
+- To get the currently-active runtime, packager, and available binaries, the user can run `volta list`.
+
+- To get *all* runtimes, packagers, and binaries currently fetched to the user's system, the user can run `volta list all`.
+
+- The user can also narrow the query:
+    - to runtimes: `volta list node`
+    - to packagers: `volta list yarn` (and, when custom installations of `npm` are supported, `volta list npm`)
+    - to a specific package: `volta list <package name>`
+    - to a specific tool: `volta list <tool name>`
```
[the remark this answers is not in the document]
chriskrycho: Ah, good catch. Updated for consistency, though depending on how we sort out the related question I may have to update it *the other direction*. 😆

text/0000-informational-commands.md
mikrostew: I think having `(defaults)` for the first is a great idea to make the distinction obvious

text/0000-informational-commands.md
mikrostew: After thinking about this more, I like the `volta list <package-or-tool>` approach, where it defaults to the package. It's more compact to type, and in cases where the binary and the package are the same name, you get more information.

text/0000-informational-commands.md
```diff
@@ -0,0 +1,766 @@
...
+## Deprecating `volta current`
+
+Since `volta list` subsumes (and substantially extends upon) the functionality of `volta current`, `volta current` should be deprecated when `volta list` is implemented. The deprecation should include a warning that the command will be removed in a future version and information about how to use `volta list --current <tool>` as a replacement.
+
+# Critique
+[critique]: #critique
+
+## Use another command name
+
+As noted in [**Why `list`?**][why-list], options for this subcommand name include `list`, `ls`, `tools`, `toolchain`, `info`, and `current`. (For discussion of shorthands, see the next section.) Given the considerable [prior art][prior-art] using `list`, however, it seems the best option in terms of what users will expect.
```
charlespierce: I like the idea (first mentioned in Discord I believe) of having `ls` be a full alias of `list`, just for ergonomic reasons. It doesn't have to be documented or shown anywhere, just as a bonus for usability & potential discoverability of the command.

text/0000-informational-commands.md
```diff
@@ -0,0 +1,766 @@
...
+
+```sh
+volta list tsc
+tool tsc / typescript@v3.4.5 node@12.2.0 npm@built-in
+tool tsc / typescript@v3.0.3 node@12.2.0 npm@built-in (default)
+```
+
+## Deprecating `volta current`
+
+Since `volta list` subsumes (and substantially extends upon) the functionality of `volta current`, `volta current` should be deprecated when `volta list` is implemented. The deprecation should include a warning that the command will be removed in a future version and information about how to use `volta list --current <tool>` as a replacement.
```
charlespierce: I like this approach a lot, it allows us to introduce the command and the deprecation warning without a breaking change, and then the next time we have a breaking change we can take that opportunity to remove the `current` command (that I suspect no one is using anyway given its limited functionality).

text/0000-informational-commands.md
```diff
@@ -0,0 +1,766 @@
...
+    - no package binaries?
+
+    And how should that be presented in `plain` vs. `human` mode? Should there be any differences between `plain` and `human` for the bare command (as in the current design)?
+
+- How should the current version be identified? It is currently marked with `(current @ <path>)`. Should this be `(from <path>)` or some other design?
+
+- Should the command accept flags to narrow the search instead of a positional argument, e.g. `volta list --node`, `volta list --yarn`, `volta list --package=typescript`, etc.?
+
+- How should the built-in version of `npm` be displayed to the user? Options include:
+    - associated with the runtime it ships with, e.g. `v12.2.0 (with npm v6.4.1)`
```
charlespierce: This is how we currently show it when you run e.g. `volta install node`, so if possible it would be good to align with that. Also, until we implement the `volta install npm` logic, listing `npm` as a separate packager seems like it has the potential for confusion, implying that the user can manage `npm` themselves.

chriskrycho:
I think this is quite ready to merge. We may continue to add more output formats and to iterate on the design as we find pain points, but the RFC has served its purpose and is fully implemented given our existing constraints from where I stand!

PR #35, opened 2019-05-23
Update RFC template to reference Notion instead of Volta

chriskrycho:
Just what it says on the tin!


chriskrycho:
Ha! Yes. Whoop.

PR #37, opened 2019-08-01
RFC: Volta Updates

charlespierce:
> Provide a path for users to keep Volta up-to-date, ranging from notifications that updates are available to automatic background updates.

[Rendered RFC](https://github.com/charlespierce/rfcs/blob/updates/text/0000-volta-updates.md)

### DIFF
--- a/text/0000-volta-updates.md
+++ b/text/0000-volta-updates.md
@@ -1,0 +1,72 @@
+- Feature Name: volta_updates
+- Start Date: 2019-08-01
+- RFC PR: (leave this empty)
+- Volta Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Provide a path for users to keep Volta up-to-date, ranging from notifications that updates are available to automatic background updates.
+
+# Motivation
+[motivation]: #motivation
+
+We are constantly working to improve Volta, so we should make sure that those updates are made available to our users, especially those who don't keep a close eye on our releases. That way users are always able to take advantage of new functionality and improvements.
+
+# Pedagogy
+[pedagogy]: #pedagogy
+
+The goal for this feature is to not require any education for typical users. The updates will be checked and made available through notifications or automatically applied, requiring nothing new from the user.
+
+For Enterprise / Advanced users that want their updates to be sourced from a private repository, we will need to provide documentation on how to direct the updater to use a non-public source, as well as how to alter any messaging so that the update notifications remain accurate for those users. This will likely need to take place in a config file within the `~/.volta` directory.
+
+# Details
+[details]: #details
+
+There are multiple approaches to updates, each building incrementally on the previous ones. This will allow us to implement straightforward approaches first, and then roll out more advanced update techniques later.
+
+## Notifications
+
+The most straightforward technique to implement would be a check of the current version against the latest available version accompanied by a notification to the user that their version is out-of-date. We can then provide a CTA that tells the user how to update (on Unix, run the install script `curl https://get.volta.sh | bash` again, or on Windows run the new installer). This still requires that the user manually update their system, but it lets the users know when that is necessary and provides guidance on how to perform that update.
+
+For Enterprise situations where Volta is installed from a different source than the default public repository, we will need to provide a way to redirect the "latest version" check to a different URL _and_ a way to alter the CTA (as these users will likely have a different install flow than the default).
+
+## Update Command
+
+After we have notifications, we should provide a command such as `volta selfupdate` that provides a single entry point to updating the user's current installation with a newer one. Implementing this will require us to have a programmatic update process that can handle any migrations from version to version that are necessary. It may be that we provide a separate "updater" application and `volta selfupdate` downloads the newest version of that and runs it. We can then change the CTA in the out-of-date notifications to simply tell the user they need to run `volta selfupdate`, instead of directing them to the specific install steps.
+
+For Enterprise users in this model, we will need to maintain the "latest version" redirect for the notification checker, but we will also need to provide a way to tell Volta where to find the updater and the associated update files, so that `volta selfupdate` works as expected.
+
+## Automatic Updates
+
+Once there is a programmatic updater that can handle the updates, the final step would be to provide an automatic update process that will keep Volta up-to-date in the background, without requiring user intervention. We will likely need a daemon to run separately and periodically check for updates, prepare the update, and implement it. Ideally we would keep the update process as atomic as possible, so that users won't be in a situation where the file structure changes but the binary hasn't yet been updated, or vice versa, which would cause transient, hard-to-diagnose bugs.
+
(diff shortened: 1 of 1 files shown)


text/0000-volta-updates.md
```diff
@@ -0,0 +1,114 @@
...
+
+- Later, we can implement an automatic updater, which runs in the background and detects when a new version is available, then downloads and unpacks the binaries, replacing the existing ones. This will result in Individual users having an experience that is effectively the same as Managed users: The binaries will be silently updated behind the scenes, without them having to take any action. This should be as atomic of a process as we can make it, so that the user is less likely to notice any strange behaviors because they are in the middle of an update.
+
+For both of these cases, we want to make sure that the checking does not have an adverse effect on the performance of the tools, especially on the performance of shims. For that reason, we likely want to perform the checks and updates in a separate process or thread, so that the main thread can continue to run. This is especially important in offline or low-bandwidth situations, where the check for new versions may take a long time before ultimately failing, and we don't want the user to have to wait.
```
dherman: We're getting into implementation details, which is probably unnecessary for this RFC, but if you're talking about doing the check in a separate thread, I would just caution that it's important to get the user-visible behavior of the checks right. You don't want even a warning to be interleaved with other user-visible behavior.

text/0000-volta-updates.md
```diff
@@ -0,0 +1,114 @@
...
+
+Additionally, for notifications, we don't want to show the notification every time the user runs a command, so we should limit the notifications to one per day so that we don't pester the user incessantly.
```
dherman: I'd also add that we maybe don't need to check for Volta updates when running shims, or at least when running their fast path, especially because you don't expect the fast path to hit the network (do I have that right? please check me on that). Maybe on the shim slow path you could still trigger the ping for new Volta versions.

text/0000-volta-updates.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Windows
+
+The Windows automatic update story is complicated, since the binaries are currently installed into the `Program Files` directory, and the user may or may not have edit access to that directory. Should we update the installer to unpack the files into the `LocalAppData` directory and then run `volta init`, in the same way that the Unix install works? That would guarantee that the user has permissions to the binaries for automatic updates, and would align the Windows and Unix stories more closely.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- Are there other update models we should implement along the way?
+- Do we want to provide separate installers? One for the "Managed environment" case and one for the "individual user" case?
```
dherman: We probably don't need a robust answer on this until and unless we have more corporate users. But maybe a simple thing we could offer for now would be tarballs/zips of the binaries, as the simplest possible thing that an IT department could use to automate the retrieval of all the Volta binaries.

rwjblue:
A few points after some discussion in this weeks volta call:

* I would like to see a little bit more information about what a "managed" installation needs to _actually_ do in order to run an update. After reading the RFC this isn't 100% obvious.
* I think that this RFC should describe a "kill switch" for the initial version. Specifically, I believe that it is **fundamentally impossible** for us to fully account for everything we _might ever_ need. My proposal here would be to have the published update bundle that is downloaded (tarballs, zip files, etc) include a file that specifically indicates "updater-protocol-version". This should be a _required_ check for both the self installed user and the managed user update scenarios.

text/0000-volta-updates.md
```diff
@@ -34,7 +34,16 @@ For Volta updates, there are three distinct sets of changes that need to be cons
 
 The large majority of updates to Volta involve _only_ updating the binaries, without requiring updates to the Data Directory or Profile.
 
-For Managed users, updating the binaries will be handled by the team managing the distribution, making sure that they are always on the latest binaries.
+### Managed Users
+
+For Managed users, updating the binaries will be handled by the team managing the distribution, making sure that they are always on the latest binaries. Specifically, we currently provide 2 binaries:
```
rwjblue: Does this mean that adding a new binary will be a "breaking change" to managed users? How do we properly communicate that a new required binary is present (or that we have removed one, if we ever decide to do that)?

I suspect we need some official metadata format that will be parsed by teams doing the managed distribution, otherwise that would always be a completely manual (and therefore error prone) process.

Thoughts?

charlespierce: I imagine we'll be distributing these in some sort of package (i.e. a tarball), so would "the contents of this archive" count as a metadata format? That's how we currently distribute the installed files for all of the Unix OS's: A tarball with all the necessary files that gets unpacked by the installer.

We could also provide some sort of "Manifest" file that lists all the binaries that are needed in order for the update to work, though I'm wary of trying to create a new format, that then will need to be translated manually into whatever format is used by the final users. Perhaps a simple list file, one file name per line, that can be published as part of the release, alongside the tarball?

rwjblue: > would "the contents of this archive" count as a metadata format?

Yes, that seems totally fine (something like "the tarball provided is unpacked directly onto the users `$PATH`). My only concern here is that (as far as I understand the system) we are **not** saying the "full tarball" is unpacked somewhere on the `$PATH` (we are saying "these 2 binaries" have to be on `$PATH`). I admit that I haven't actually dug into the contents of that tarball so maybe I'm just missing something silly here.

> Perhaps a simple list file, one file name per line, that can be published as part of the release, alongside the tarball?

Ya, this seems good to me also, though I think it would be perfectly fine if this `BinaryFiles.txt` file (or whatever we call it 😛) is actually **inside** the root of the tarball. That makes it pretty easy to bundle everything together, and it makes it a bit clearer what the filenames are "relative to".

charlespierce: Currently the tarball includes the 2 binary files (`volta` and `shim`) as well as the two shell-specific function wrappers (`load.sh` and `load.fish`). With the shift to basing everything on the binaries themselves, the latter two would be going away, so the tarball would actually only include the necessary binary files.

That said, it's straightforward enough to also create / include a manifest file of some kind inside the tarball, so I'm open to either solution.

rwjblue: > Currently the tarball includes the 2 binary files (volta and shim) as well as the two shell-specific function wrappers (load.sh and load.fish).

I see! I'm sorry, I didn't realize that was all that the current tarball includes. If we think that will continue to be the case, we can definitely say "all of the tarball contents get extracted to somewhere on the users $PATH", but I _suspect_ that we might also want to include some other files eventually (e.g. `README.md`, `LICENSE`, etc) so if the manifest route isn't _too_ bad maybe we should still do it?

charlespierce: Makes sense to me, and it shouldn't be bad at all, since it will change very rarely. We can create a `volta.manifest` file and include it in the tarball, then provide documentation that indicates what needs to be included in an "install" are only the files listed in the manifest.

PR #38, opened 2019-09-11
Road to 1.0 Release

charlespierce:
> Document the features necessary to release Volta 1.0

[Rendered RFC](https://github.com/charlespierce/rfcs/blob/road_to_10/text/0000-road-to-10.md)

### DIFF
--- a/text/0000-road-to-10.md
+++ b/text/0000-road-to-10.md
@@ -38,6 +38,10 @@ We currently don't support Pinning or Installing a different version of `npm` th
 
 For advanced or debugging use-cases, we should support a `volta run` command that allows users to specify the versions of Node, npm, and Yarn to use at the command line. This will allow users to override the platform detected by Volta, in cases where they explicitly want to pick a different version of Node. See https://github.com/volta-cli/volta/issues/546
 
+## Yarn 2.0 / Plug'n'Play
+
+Yarn recently released version 2.0, which comes with a number of structural changes to how Yarn works. The largest of those changes is that Yarn Plug'n'Play is the behavior by default, so we will need to ensure support for that in order to make Yarn 2.0 work as expected with Volta. See https://github.com/volta-cli/volta/issues/651
+
 ## Documentation
 
 As part of working towards 1.0, we should do a pass on our documentation and make sure that everything is up-to-date and accurate. We should also make sure that there aren't any major features that are missing documentation. This will improve the onboarding experience and allow new users to more easily discover how they can use Volta.
@@ -46,7 +50,6 @@ As part of working towards 1.0, we should do a pass on our documentation and mak
 
 There are a number of issues that are relatively small on their own, but combined represent a significant amount of work. Each of them represents polish that will make Volta more delightful to use. For example (not an exhaustive list):
 
-* https://github.com/volta-cli/volta/issues/535
 * https://github.com/volta-cli/volta/issues/523
 * https://github.com/volta-cli/volta/issues/509
 * https://github.com/volta-cli/volta/issues/495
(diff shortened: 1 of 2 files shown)


text/0000-road-to-10.md
```diff
@@ -57,7 +57,11 @@ There are a number of issues that are relatively small on their own, but combine
...
 
-Another feature that could be considered for 1.0 is support for Yarn Plug'n'Play, which installs all of the dependencies from a local cache and doesn't use `node_modules`. However, we haven't yet received requests to support that functionality, and it will require a fair amount of investigation on the implementation. One major issue is that Yarn's feature relies on JS Monkeypatching, which we can't take advantage of in Rust, so we would need to either call Node (which will have performance implications) or duplicate the Yarn logic within Volta (which will take significant development effort).
+A feature that could be considered for 1.0 is support for Yarn Plug'n'Play, which installs all of the dependencies from a local cache and doesn't use `node_modules`. However, we haven't yet received requests to support that functionality, and it will require a fair amount of investigation on the implementation. One major issue is that Yarn's feature relies on JS Monkeypatching, which we can't take advantage of in Rust, so we would need to either call Node (which will have performance implications) or duplicate the Yarn logic within Volta (which will take significant development effort).
```
chriskrycho: Note that while we haven't *yet* hit this, some puzzle pieces *should* 🤞 fall into place in the Ember community (many of our early adopters) over the next few months that unlock Yarn P'n'P there, making adoption for some of our users much likelier.

charlespierce: Good to know :+1: We'll definitely need to keep an eye on that and bump it up in priority if it becomes a requested feature.

dherman:
I suspect we can remove PnP support from the list of 1.0 blockers. So far it seems like PnP has enough adoption issues that it's not being widely used yet and may evolve.

rwjblue:
> I suspect we can remove PnP support from the list of 1.0 blockers.

Ya, seems fine to me. It also seems like something we can do within the 1.x timeframe without any backwards incompatible changes.

text/0000-road-to-10.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Apple Silicon Support
+
+The new Apple processors are available in laptops and showing lots of promise. Though we should support native builds on those processors, currently Volta works correctly under Rosetta 2, so this isn't a blocker for a 1.0 release. When we do add support, we also need to take into account that we may need to fall back to downloading x64 versions of Node on Apple devices, since there may not be native builds available for older versions of Node.
+
+## pnpm Support
+
+Currently, Volta only supports npm and Yarn for package managers, as they are the two most popular. [pnpm](https://pnpm.js.org/) is a relative newcomer to the space, however it is growing in adoption and we have received a few requests to support it in Volta. However, there are some technical hurdles (most notably pnpm's clobbering of `NODE_PATH` when running executables) that make it difficult to implement at the moment. While pnpm support is definitely on the near-term roadmap (likely one of the first features to be tackled after 1.0), the volume of users requesting `pnpm` is low enough that it isn't necessary as part of the 1.0 feature set.
```
dherman, line 57: A couple slight verbiage suggestions to really emphasize that we care about supporting pnpm:

> However, there are some technical hurdles

I'd change this to "There are a few remaining technical hurdles (most notably etc etc) blocking the implementation at the moment" so it doesn't sound like we have doubts about whether we _can_ do it, just a few steps left to go, with a sense of inevitability that we'll get there.

> the volume of users requesting `pnpm` is low enough that it isn't necessary as part of the 1.0 feature set

This is a little unnecessarily dismissive. Maybe something like "While pnpm support is definitely on the near-term roadmap (likely one of etc etc), the core of Volta has stabilized to the point that we don't need to block shipping for the large number of users of other package managers."

PR #39, opened 2019-09-14
Command for overriding platform: `volta run`

charlespierce:
> Add a `volta run` command, allowing users to execute tools with an override on the detected platform.

[Rendered](https://github.com/charlespierce/rfcs/blob/volta_run/text/0000-volta-run.md)

### DIFF
--- a/text/0000-volta-run.md
+++ b/text/0000-volta-run.md
@@ -16,7 +16,7 @@ Volta currently will automatically detect the platform (Node runtime and yarn ve
 # Pedagogy
 [pedagogy]: #pedagogy
 
-We will need to document the existence of `volta run` and how the users can specify the versions of Node and Yarn when running a command. This will be provided on the command line help output, however it should also be explained in detail in the documentation. We should also document how the overall platform is resolved if the user specifies only one of Node or Yarn, so that the behavior is clear to users.
+We will need to document the existence of `volta run` and how the users can specify the versions of Node and Yarn when running a command. This will be provided on the command line help output, however it should also be explained in detail in the documentation. We should also document on the website how the overall platform is resolved if the user specifies only one of Node or Yarn, so that the behavior is clear to users.
 
 # Details
 [details]: #details
@@ -27,7 +27,15 @@ The new command will have the structure:
 volta run [--node <version>] [--yarn <version>] <command>
 ```
 
-Specifically, the `--node` and `--yarn` settings will need to be specified _before_ the command to be executed, as everything after the command will be treated as arguments to the command itself. This will allow users to specify a full command, with any command-line flags they need, without having to append `--` to separate the `volta run` command from the others.
+For example:
+
+```
+volta run --node 8 --yarn 1.10 ember new my-app --yarn
+```
+
+Would run the command `ember new my-app --yarn` with Node 8 and Yarn 1.10.
+
+Of note, the `--node` and `--yarn` settings will need to be specified _before_ the command to be executed, as everything after the command will be treated as arguments to the command itself. This will allow users to specify a full command, with any command-line flags they need, without having to append `--` to separate the `volta run` command from the others.
 
 ## Platform Resolution
 
@@ -41,11 +49,15 @@ We will resolve the a platform that has `node@8` and `yarn@1.17.3`, with the lat
 
 ## 3rd-Party Binaries
 
-After resolving the platform, if the command the user is executing is a 3rd party binary (not `node`, `npm`, `npx`, or `yarn`), then we will still do resolution of that binary relative to the current project (if any). That means if we are inside a project with that binary as a dependency, we will still use the project-local version of that tool.
+After resolving the platform, if the command the user is executing is a 3rd party binary (not `node`, `npm`, `npx`, or `yarn`), then we will need to maintain the existing resolution of that binary relative to the current project (if any). That means if we are inside a project with that binary as a dependency, we will still use the project-local version of that tool, however we will use the Node / Yarn versions specified on the command-line to execute the tool.
 
 ## Verbose Output
 
 In order to provide meaningful verbose output, we will need to update the `SourcedPlatformSpec` model to support platforms that come from the command line as well. This will likely require a refactor to support specifying the version for each tool explicitly, as opposed to having a single `Source` for the Platform as a whole.
+
+## Additional Package Managers
+
+Currently, the only package manager that we support specifying on its own is Yarn. However, there are plans to support `npm` in the near future and farther out, we may want to support others (e.g. `pnpm`, `entropic`, etc). For each aspect of the platform that we support, we'll need to add an accompanying flag to the `run` command that supports specifying the version of that tool.
 
 # Critique
 [critique]: #critique
@@ -68,5 +80,5 @@ However, it would also mean that if a user doesn't specify `--node`, that no com
 # Unresolved questions
 [unresolved]: #unresolved-questions
 
-- Do we want to support removing `yarn` from the platform (e.g. `--yarn none`) in a way that won't be overridden by the platform inheritance?
+- Do we want to support removing `yarn` from the platform (e.g. `--no-yarn`) in a way that won't be overridden by the platform inheritance?
 - Do we want to support changing a package version as well, letting the user run a different version of a tool than they currently have installed / available?
--- a/text/0000-volta-run.md
+++ b/text/0000-volta-run.md
@@ -24,10 +24,10 @@ We will need to document the existence of `volta run` and how the users can spec
 The new command will have the structure:
 
 ```
-volta run [--node <version>] [--yarn <version>] <command>
+volta run [--node <version>] [--yarn <version> | --no-yarn] [--env <variable>=<value>] <command>
 ```
 
-For example:
+Some examples:
 
 ```
 volta run --node 8 --yarn 1.10 ember new my-app --yarn
@@ -35,7 +35,13 @@ volta run --node 8 --yarn 1.10 ember new my-app --yarn
 
 Would run the command `ember new my-app --yarn` with Node 8 and Yarn 1.10.
 
-Of note, the `--node` and `--yarn` settings will need to be specified _before_ the command to be executed, as everything after the command will be treated as arguments to the command itself. This will allow users to specify a full command, with any command-line flags they need, without having to append `--` to separate the `volta run` command from the others.
+```
+volta run --node latest --env EMBER_ENV=production ember test
+```
+
+Would run the command `ember test` with the latest Node and with the environment variable `EMBER_ENV` set to `production`.
+
+Of note, all of the flags will need to be specified _before_ the command to be executed, as everything after the command will be treated as arguments to the command itself. This will allow users to specify a full command, with any command-line flags they need, without having to append `--` to separate the `volta run` command from the others.
 
 ## Platform Resolution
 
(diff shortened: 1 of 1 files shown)


text/0000-volta-run.md
```diff
@@ -0,0 +1,72 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+Volta currently will automatically detect the platform (Node runtime and yarn version) based on the context of where you are in the file system. However, there are occasionally times when you want to run a command using a different version of Node or Yarn than what Volta detects, without changing the pinned or default versions. To support those use-cases, we should provide a command that allows users to execute a command with a platform specified on the command-line.
+
+# Pedagogy
+[pedagogy]: #pedagogy
+
+We will need to document the existence of `volta run` and how the users can specify the versions of Node and Yarn when running a command. This will be provided on the command line by StructOpt, however it should also be explained in detail in the documentation. We should also document how the overall platform is resolved if the user specifies only one of Node or Yarn, so that the behavior is clear to users.
```
rwjblue:
```suggestion
We will need to document the existence of `volta run` and how the users can specify the versions of Node and Yarn when running a command. This will be provided on the command line help output, however it should also be explained in detail in the documentation. We should also document how the overall platform is resolved if the user specifies only one of Node or Yarn, so that the behavior is clear to users.
```

`StructOpt` seems like a _somewhat_ unimportant technical detail here, and I think the userland thing that you meant was that the `volta` usage output would include info on `run`.

rwjblue: > We should also document how the overall platform is resolved if the user specifies only one of Node or Yarn, so that the behavior is clear to users.

Do you mean in the website? In the help output? Both?

text/0000-volta-run.md
```diff
@@ -0,0 +1,72 @@
...
+
+# Details
+[details]: #details
+
+The new command will have the structure:
+
+```
+volta run [--node <version>] [--yarn <version>] <command>
```
rwjblue: Since we ultimately _plan_ to allow custom `npm` version pinning, we should at least consider (here, in unanswered questions, a future changes section) what pinning the `npm` version would look like also.

In addition, there are quite a few _other_ tools / registries that we _might_ eventually want to support (entropic, pnpm, etc). Do we need an option for each?

rwjblue: Can you add an example command here (just below the usage)? It is not obvious how the command is specified and we don't give an example until a few sections later.

text/0000-volta-run.md
```diff
@@ -0,0 +1,72 @@
...
+```
+
+Specifically, the `--node` and `--yarn` settings will need to be specified _before_ the command to be executed, as everything after the command will be treated as arguments to the command itself. This will allow users to specify a full command, with any command-line flags they need, without having to append `--` to separate the `volta run` command from the others.
```
rwjblue: I think it's definitely great to not _require_ the `--` separator, but I'd like to suggest that we consider supporting it if used. The `--` as a separator is _pretty_ common, and might feel more natural to folks. I'm a bit torn there because this may just make it more complicated (having to display both versions in docs and stuff) and choosing only one may be best. 

What do you think?

text/0000-volta-run.md
```diff
@@ -0,0 +1,72 @@
...
+
+```
+volta run --node 8 ember serve
+```
+
+We will resolve the a platform that has `node@8` and `yarn@1.17.3`, with the latter pulled from the existing pinned project and the former overridden by the command-line flag. This allows users to specify the parts of the platform that they want to change, while avoiding the necessity of specifying everything, even those things that don't change.
+
+## 3rd-Party Binaries
+
+After resolving the platform, if the command the user is executing is a 3rd party binary (not `node`, `npm`, `npx`, or `yarn`), then we will still do resolution of that binary relative to the current project (if any). That means if we are inside a project with that binary as a dependency, we will still use the project-local version of that tool.
```
rwjblue: Can you tweak the prose here to make it clearer that:

1) this isnt' a behavior change
2) that the tool (local or global) will specifically use the node/yarn version specified

I _think_ these are both true, but the prose doesn't make it clear when I first read...

text/0000-volta-run.md
rwjblue: Some examples that might be useful here:

```
volta run --node 10 npm run some-script -- --somescript-arg=true
volta run --node 8 --yarn 1.17 yarn some-script --somescript-arg
volta run --node 12 ember build
```

text/0000-volta-run.md
```diff
@@ -0,0 +1,72 @@
...
+There are alternative names we could use for the command, such as:
+
+- `volta exec`
+- `volta do`
+
+However, `volta run` aligns with the "run script" command in both `npm` and `yarn`, so it seems to fit the best for this use-case.
+
+## Platform Resolution
+
+Instead of inheriting the platform from the current, we could treat it as purely defined on the command-line. This would give users total control over all of the pieces of the platform, including allowing them to remove `yarn`, even if it is specified in the project or default platforms.
```
rwjblue: I'd prefer to do `--no-yarn` style option to support the "don't fallback" style.

text/0000-volta-run.md
charlespierce: Good question! My intention was on the website, where we can be more long-form about the nuances of the command.

text/0000-volta-run.md
charlespierce: Done and done. I added an example immediately after the usage, and a new section "Additional Package Managers" to cover how it would look as we add support for other parts of the platform.

text/0000-volta-run.md
charlespierce: I'm not 100% sure, but I think that we will implicitly support the `--`, however we'll still need the `--node`, `--yarn`, etc. to come before the command.

text/0000-volta-run.md
charlespierce: Good catch, cleaned that up a bit.

text/0000-volta-run.md
charlespierce: I like that better than the example I had in "Unresolved Questions" of `--yarn none`. Still open if we want to support that, but I can see it being useful.

charlespierce:
@rwjblue Thanks for the feedback, that's great and super helpful! To answer your other questions:

1. Environment variables can be specified in the usual way `VARIABLE=value volta run ...` on Unix, though I could see the benefit of having a `--env VARIABLE=value` flag as well, to unify the UX and support those on Windows (which doesn't have an easy shorthand for setting one-off environment variables in all shells).

2. We still need to parse the first part of the command to figure out which shim behavior to use (in the same way that we parse the command name when running `node` or `ember`). The arguments, however, are just in an array and will be left untouched.

3. See the answer for 1, you can / should specify them before `volta run`

4. We don't have to specialize `--node` and `--yarn`, we could probably do something like `--with node@8`. My initial concern when doing that was how would we handle if someone included the same tool multiple times (e.g. `--with node@8 --with node@10`, but in those cases we can probably error out early and tell them to only set each value once.

5. I like the idea of supporting a JSON config, though specifying it directly on the command line seems tricky, since in order for it to be detected as a single argument, it will need to be quoted. That combined with necessary quotes around the property names could get unwieldy for the user. But supporting a JSON config file should be relatively straightforward.
    - One thing to note for that, however, is that currently the JSON objects that we deal with represent fully-resolved versions (i.e. You can't have `"node": "8"` in your `package.json`). On the command-line, we would want to support partially-specified versions, so that will be a slight departure in this use-case.

I'll think on some of these issues a bit more and then look to fill out some more details :+1:

text/0000-volta-run.md
rwjblue: @charlespierce - The main thing I was pointing out here is that the `--` in the first example was how `npm` has you pass arguments/options to a `npm run-script`.

rwjblue:
> which doesn't have an easy shorthand for setting one-off environment variables in all shells

Great point!

> We still need to parse the first part of the command to figure out which shim behavior to use (in the same way that we parse the command name when running node or ember). 

From initially reading this I would assume that `<command>` is _literally_ what I would have typed on the command line to run the command. This lead me to believe that the following would work (and was being proposed by this RFC):

```
volta run --node 10 EMBER_ENV=production ember build
```

However, I don't _think_ thats correct (at least it doesn't seem like thats what you were saying in your comment above). Can you confirm/clarify?

charlespierce:
To clarify, that won't work with the current proposal. Ultimately, we have to do _some_ parsing of the passed-in command, in order to figure out which tool is being executed and run our unique logic for that tool. Since we can't reasonably reproduce the command-parse logic of every shell, we need the command to start with the name of the tool (in your example, we would see `EMBER_ENV=production` as the "tool name" and then fail to find that "tool").

I wonder if there's a better way we can explain / clarify how that works (and maybe having the `--env` option would reduce the chance of a user accidentally doing that.

rwjblue:
> Since we can't reasonably reproduce the command-parse logic of every shell, we need the command to start with the name of the tool

Yeah, I definitely see the issue you are describing (this is _exactly_ the reason I asked about parsing above). I am definitely liking the idea of passing the env vars via `--env` to `volta run` (I've been looking for a convenient cross platform way of doing this sort of thing for my CI runs in various projects). If we do that **and** essentially "lint" the string after known volta arguments to emit a nice warning when you try what I had pasted (`volta run --node 8 EMBER_ENV=foo ember b`) it may be good enough...

--- 

Oh, also this question:

> Have we considered supporting environment variables (in addition to or instead of volta run)?

Was talking about `VOLTA_NODE=8 ember b` (not like the third question which was about passing environment variables to the invoked command).

charlespierce:
Added a bunch more to address many of the comments raised by @rwjblue

text/0000-volta-run.md
```diff
@@ -59,14 +85,43 @@ There are alternative names we could use for the command, such as:
 
 However, `volta run` aligns with the "run script" command in both `npm` and `yarn`, so it seems to fit the best for this use-case.
 
+## Version Flags
```
rwjblue: Thank you for adding this section. I do agree with the conclusion (that the more common case should be what we optimize), and I think that as written this RFC doesn't _prevent_ us from adding `--with <any valid volta install argument>` in the future as a follow up RFC.

👍

text/0000-volta-run.md
```diff
@@ -59,14 +85,43 @@ There are alternative names we could use for the command, such as:
...
+
+- It requires users to think about the existence of Volta when running the shims. When a user calls `yarn`, the goal of Volta is that they shouldn't have to think about the fact that Volta is figuring it out, it should just work as they expect. If we add environment variables that can change how those work, it could lead to difficult-to-debug situations where a user has an exported variable in their terminal that is breaking things and it's not immediately obvious why.
+
+- It's not as ergonomic cross-platform. As mentioned above, Windows doesn't have as easy as shorthand for setting environment variables on a given command, so Windows users would have to jump through more hoops to get access to this behavior. Volta is committed to keeping Windows as a first-class citizen, so we don't want to break that by relying too much on Unix-isms.
```
rwjblue: IMHO, this is the deal breaker for this one.

text/0000-volta-run.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Verbose Output
+
+In order to provide meaningful verbose output, we will need to update the `SourcedPlatformSpec` model to support platforms that come from the command line as well. This will likely require a refactor to support specifying the version for each tool explicitly, as opposed to having a single `Source` for the Platform as a whole.
+
+## Additional Package Managers
+
+Currently, the only package manager that we support specifying on its own is Yarn. However, there are plans to support `npm` in the near future and farther out, we may want to support others (e.g. `pnpm`, `entropic`, etc). For each aspect of the platform that we support, we'll need to accompanying flags to the `run` command that support specifying the version of that tool. (i.e. `--pnpm <version>` and `--no-pnpm`).
```
rwjblue:
```suggestion
Currently, the only package manager that we support specifying on its own is Yarn. However, there are plans to support customizing the `npm` version independently from the version installed by node in the near future and farther out, we may want to support others (e.g. `pnpm`, `entropic`, etc). For each aspect of the platform that we support, we'll need to accompanying flags to the `run` command that support specifying the version of that tool. (i.e. `--pnpm <version>` and `--no-pnpm`).
```

text/0000-volta-run.md
charlespierce: Definitely not, we can always make changes in a follow-up, or if we decide after actually using the command that it needs to change.

text/0000-volta-run.md
charlespierce: Agreed, especially as we’re already supporting the `--env` option in the other use-case.

PR #40, opened 2019-10-14
Installing/pinning npm

charlespierce:
> Support installation and pinning of arbitrary `npm` versions, decoupled from the selected Node version.

[Rendered RFC](https://github.com/charlespierce/rfcs/blob/install_pin_npm/text/0000-install-pin-npm.md)

### DIFF
--- a/text/0000-install-pin-npm.md
+++ b/text/0000-install-pin-npm.md
@@ -16,50 +16,55 @@ For package managers, we currently only support using an arbitrary version of `y
 # Pedagogy
 [pedagogy]: #pedagogy
 
-There shouldn't be anything new we need to document / teach users, as the behavior already exists for Node and Yarn. The only change is that we will support the `volta install npm` and `volta pin npm` commands in the same way that we support `volta install yarn` and `volta pin yarn`.
+The main concept we will need to document is the concept of bundled versions, and how they differ from explicit versions within the Volta model. The hope is that this is intuitive, however we should document the potential ambiguities so that users can easily discover how to get Volta to do what they want.
+
+Beyond that concept, all of the other behavior is already provided in our implementation of Node and Yarn. The only change is that we will now support `volta install npm` and `volta pin npm` in addition to the other tools.
 
 # Details
 [details]: #details
 
+## Bundled vs. Explicit Versions
+
+One point of ambiguity when dealing with `npm` is that Node always comes bundled with a version of `npm`, so whenever a user upgrades their version of Node, they are also upgrading the bundled version of `npm`. This can lead to a situation where the user has picked a version of `npm` but is now upgrading to a Node version that has a newer version of `npm` available. To resolve this potential ambiguity, this RFC proposes being clear about the difference between a bundled version and an explicitly chosen version.
+
+If the user has not specified (via `volta pin` or `volta install`) a version of `npm`, then we write nothing into the respective `json` files, and when we run `npm`, we run the bundled version. This means if the user hasn't specified `npm` and upgrades their version of Node, their `npm` install will also be upgraded. Currently, this is always the behavior of `npm`, so we preserve that for those users who don't want or need to specify a custom version.
+
+If, however, the user has explicitly chosen a version of `npm`, then we write that into the appropriate `json` file and we always respect the user's choice. If the user upgrades to a version of Node with a newer version of `npm`, until they explicitly also upgrade to a newer version of `npm`, we maintain their previous setting. We should, however, show an informational message to the user, letting them know that the bundled version includes a newer `npm` and providing them with a command they can use to switch to the bundled version.
+
 ## Installation
 
-Unlike Yarn, `npm` doesn't provide a pre-packaged tarball for download. Instead, the rely on using the existing `npm` installation that is provided with Node to do any upgrades. This is very similar to how we handle global 3rd-party packages, so we should be able to leverage a similar approach for installing `npm`. That does mean, however, that in the same way we require Node to be available to install packages, we will also require Node to be available to install a custom version of `npm`.
+Similar to Yarn, we can download the `npm` tarball directly from `https://npmjs.org/npm/-/npm-{{version}}.tgz`, which comes pre-packaged with all of the necessary dependencies already included. This means that we can treat npm exactly the same way we treat Yarn, unpacking it into its own directory. The only issue is that the executable scripts provided by the tarball have extra path-finding behaviors that don't work with Volta's model, so those will need to be overwritten (see below for details).
 
-It appears we can also skip the `npm install` step, as the `npm` package tarball comes with all of the dependencies already included.
+To support downloading from custom repositories, we will also need to support hooks for `npm`, allowing the user to redirect the `index`, `latest`, and `distro` checks to a different URL. This mirrors the implementation of Yarn within Volta.
+
+Additionally, we should provide a way for the user to switch from an explicit version of `npm` (specified in the `package.json` file) to the bundled version (with no entry in the `package.json`). To support this, we can allow a custom "version" that represents the bundled `npm`, for example "bundled", meaning the user can switch to use the bundled version with `volta pin npm@bundled`. To help make this custom version more discoverable, the error message a user sees when they run `volta uninstall npm` should include a CTA to run `volta install npm@bundled` or `volta pin npm@bundled` if they want to switch to use the bundled version and stop using an explicit version.
 
 ## Running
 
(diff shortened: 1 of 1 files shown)


text/0000-install-pin-npm.md
```diff
@@ -0,0 +1,65 @@
...
+There shouldn't be anything new we need to document / teach users, as the behavior already exists for Node and Yarn. The only change is that we will support the `volta install npm` and `volta pin npm` commands in the same way that we support `volta install yarn` and `volta pin yarn`.
+
+# Details
+[details]: #details
+
+## Installation
+
+Unlike Yarn, `npm` doesn't provide a pre-packaged tarball for download. Instead, the rely on using the existing `npm` installation that is provided with Node to do any upgrades. This is very similar to how we handle global 3rd-party packages, so we should be able to leverage a similar approach for installing `npm`. That does mean, however, that in the same way we require Node to be available to install packages, we will also require Node to be available to install a custom version of `npm`.
+
+It appears we can also skip the `npm install` step, as the `npm` package tarball comes with all of the dependencies already included.
```
rwjblue: If we skip the `npm install` step, what are we using the global node installation for?

text/0000-install-pin-npm.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Running
+
+Currently, to execute `npm`, we add the Node bin directory to the PATH and then call `npm`, relying on the fact that `npm` is bundled with Node, so `npm` is available within the Node bin directory. To support arbitrary versions of `npm`, we will need to take the approach that is used for packages and call the `npm-cli.js` file directly. Since this isn't an arbitrary package, but will always be `npm`, we can likely shortcut the configuration files and detection of binaries that we need for packages and use the known path to the file.
```
rwjblue: Does this mean that we will defer to the "project local" version if the project depends on `npm`? That seems at the same time "natural" and "weird" 😝

text/0000-install-pin-npm.md
```diff
@@ -0,0 +1,65 @@
...
+
+For Windows compatibility, we will need to call `node /path/to/npm-cli.js` instead of executing the `.js` file directly, since Windows doesn't understand shebang loaders.
```
rwjblue: Any reason not to do this across the board?

text/0000-install-pin-npm.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Volta Directory Layout
+
+We currently unpack Node in `tools/image/node/{node_version}/{bundled_npm_version}`. However, the final directory representing the bundled `npm` version isn't needed, since we will be keeping each version of `npm` separate from the `node` image directory anyway. For simplicity, we should remove the extra directory and install Node under `tools/image/node/{node_version}`. That will allow us to undo some of the coupling between Node and `npm` that currently exists in the code base, as well as more easily support arbitrary versions of `npm`.
+
+The `npm` packages themselves should be unpacked under `tools/image/npm/{npm_version}`, in the same way that `yarn` is structured currently.
+
+## Upgrading Node
+
+One tricky point with supporting `npm` is that it is always bundled with Node. This means that there is some potential ambiguity when a user upgrades their version of Node: If the version of `npm` that comes with that Node is newer than the one they have pinned, do they want to upgrade to use the bundled version, or stay with the pinned version? There isn't a clear default choice that makes sense, so instead in that situation we can ask the user:
```
rwjblue: I like this, but I have historically had pretty significant issues testing command line tools that have interactive prompting like this. Do we currently have any "prompting"? It might be a good idea to decouple from the prompt (and add it as an improvement after landing the "MVP").

One possible way to avoid prompting, would be something like:

```bash
$ volta pin node@latest
...
node@12.12.0 includes a newer version of npm (6.11.3) than you currently have pinned in this project (6.9.0)! In order to update your pinned version of npm, please run:

  volta pin npm@6.11.3
```

text/0000-install-pin-npm.md
```diff
@@ -0,0 +1,65 @@
...
+
+```bash
+$ volta pin node@latest
+...
+node@12.12.0 includes a newer version of npm (6.11.3) than you currently have pinned in this project (6.9.0)!
+Would you like to upgrade to npm@6.11.3 (Y/n)?
+```
+
+If the user chooses to upgrade to the bundled version, we can remove the custom setting and allow the project to use the default behavior of the bundled `npm`. If they choose not to upgrade, we leave the custom setting as-is.
```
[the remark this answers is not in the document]
charlespierce: I think this ties into the above suggestion. If we _do_ prompt and the user chooses to upgrade, then we need to change the `package.json` otherwise the user's decision to upgrade doesn't mean anything.

On the flip-side, if we default to keeping the pinned version as-is, with a warning, then I agree we shouldn't make any changes to `package.json` when altering Node.

text/0000-install-pin-npm.md
charlespierce: We don't currently have any prompting, no, so that would be new behavior for Volta. I think I like the idea of not making any changes, but providing a message that lets the user know there's a newer version and how to upgrade. I especially like the suggestion of including the exact command the user needs to run in order to perform the upgrade, that's a great UX!

text/0000-install-pin-npm.md
charlespierce: That actually was what I was thinking as I wrote it, but no, we can definitely do that for both cases. However, one thing I realized considering this is that we'll likely still need to include the `npm` directory on the PATH, and will probably need to do some rigorous testing to account for the situations where an `npm` or `yarn` script calls another `npm` script, where we will need to make sure the command `npm` executes correctly in the nested case.

text/0000-install-pin-npm.md
charlespierce: Probably not, though we could look at doing so. My thought was that the logic would be similar, but still different enough to be distinct, so they would have different behavior. Specifically, they would be a sort of melding of the logic for `yarn` and the logic for packages.

text/0000-install-pin-npm.md
charlespierce: We're using the global node installation (specifically the bundled `npm` in the global node installation) for resolving the version using `npm view` and fetching the tarball using `npm pack`

Given the lack of an `npm install` step, we could potentially take an approach that more closely mimics `yarn` and builds the distro index and download URL directly from `npmjs.com`, with it's own hooks to support changing the download URL (in the same way that Yarn does).

text/0000-install-pin-npm.md
```diff
@@ -0,0 +1,65 @@
...
+[pedagogy]: #pedagogy
+
+There shouldn't be anything new we need to document / teach users, as the behavior already exists for Node and Yarn. The only change is that we will support the `volta install npm` and `volta pin npm` commands in the same way that we support `volta install yarn` and `volta pin yarn`.
+
+# Details
+[details]: #details
+
+## Installation
+
+Unlike Yarn, `npm` doesn't provide a pre-packaged tarball for download. Instead, the rely on using the existing `npm` installation that is provided with Node to do any upgrades. This is very similar to how we handle global 3rd-party packages, so we should be able to leverage a similar approach for installing `npm`. That does mean, however, that in the same way we require Node to be available to install packages, we will also require Node to be available to install a custom version of `npm`.
```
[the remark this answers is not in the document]
charlespierce: Correct, the goal would be to be exactly the same as `yarn`: Using `npm` in a pinned package will work as it always has. Calling `volta pin npm` will only work if at least `node` is already pinned (we currently will throw an error if you try to `volta pin yarn` before pinning Node, we would do the same for npm).

Calling `volta install npm` under this suggestion would require you to have already run `volta install node`, which diverges from Yarn, however I think that's a bug in the Yarn behavior moreso than this proposed `npm` behavior (see volta-cli/volta#575).

charlespierce:
Thanks for the comments @rwjblue, they definitely helped clarify the thinking / plans on several points. I'll update the RFC soon with changes to reflect the improved process.

To answer your other questions:

1. The behavior of `volta install node` wouldn't change, you would still have `npm` available because it was bundled with `node` (unless you had previously selected a specific version with `volta install npm`, then you would retain that specific version).
2. The "default installed npm" would be special-cased, in that it would be treated as "no custom npm". We already have logic for that in the data structures (`volta` key in `package.json`, etc.).
    * If the pinned version of npm exactly matches the bundled default version (which it currently always will), then we treat it the same as if no version is specified.
    * If the pinned version is _different_ from the bundled default version (after this RFC), then we would create a `tools/image/npm/{npm_version}` directory for it.
    * This could potentially lead to some duplication, if you install a newer version of npm and then upgrade your node to one that has that version pinned, you will have 2 copies of it, however since `npm` is always bundled with Node anyway, there's no real way to avoid that duplication.
3. I'll add a mention of it, but `volta list` would include `npm` in the list of package managers, along-side `Yarn`. I believe @chriskrycho's architecture of that command can already handle `npm` with little adjustment needed, so it should be a smooth migration on that front.

rwjblue:
Awesome, thank you

charlespierce:
@rwjblue Updated based on the above discussions and further thought, thanks again!

PR #43, opened 2020-04-24
Support for Monorepos / Workspaces

charlespierce:
> Improve the ergonomics of working with Volta in a monorepo environment.

[Rendered RFC](https://github.com/charlespierce/rfcs/blob/monorepos/text/0000-monorepo-support.md)

### DIFF
--- a/text/0000-monorepo-support.md
+++ b/text/0000-monorepo-support.md
@@ -1,0 +1,81 @@
+- Feature Name: monorepo_support
+- Start Date: 2020-04-23
+- RFC PR: (leave this empty)
+- Volta Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Improve the ergonomics of working with Volta in a monorepo environment.
+
+# Motivation
+[motivation]: #motivation
+
+Currently, Volta treats the first `package.json` file it finds in the directory tree as the root of the current project and doesn't look any further for more context. This means that users in a monorepo environment are forced to duplicate their Volta settings in each subproject. Additionally, if they ever want to update their pinned `node` or `yarn` versions, they need to update each subproject individually, negating many of the reasons for using a monorepo in the first place. Volta should provide more ergonomic solutions so users in a monorepo can manage their tool versions in a single place without having to duplicate the settings in each subproject.
+
+# Pedagogy
+[pedagogy]: #pedagogy
+
+The changes proposed by this RFC are primarily advanced features required by users who _manage_ monorepo projects. Users who are purely consumers should be able to continue using Volta without needing to consider the details of how Volta determines the appropriate context. For the advanced users, we will need to document:
+
+- Volta's model for having multiple project roots, using the `"extends"` key in `package.json`
+- The logic Volta uses to combine platform settings from multiple locations into a single context (including the ability to disable inheriting with the `false` value).
+- The additional impact of having multiple project roots, i.e. `hooks` and local binary resolution.
+- The behavior of `volta pin` in subprojects vs the root project.
+
+Ideally, we should document all of these with specific examples to show how the different pieces interact. We could potentially even create an `examples` repo (or directory within the main repo) for users to reference.
+
+# Details
+[details]: #details
+
+## Multiple Roots
+
+The core change to Volta will be expanding our `Project` model to support a _sequence_ of project roots, instead of a single project root. We will support a sequence by adding the ability to put an `"extends"` entry inside the `"volta"` section of `package.json`. The value of that entry will be the path to another json file, which we will parse just like the original `package.json`. The directory that new file is in will be the next value in the sequence of project roots. Then, if that new file also has an `"extends"` entry, we will continue the same process, building up effectively a linked list of project roots to use for all of our project information.
+
+Importantly, the order of the sequence of project roots will start with the closest `package.json` (what we previously called the sole project root), and then follow the `"extends"` chain out from there.
+
+## Platform Resolution
+
+To resolve the Project platform, taking into account all of the settings in the different project files, we will merge the platform settings from each project file, giving precedence at all times to the files earlier in the sequence. So the highest precedence will be at the existing project root, and then each file that we resolve after that will only update settings that haven't been specified by an earlier entry.
+
+To support a use-case where a project wants to exclude e.g. `yarn` and _not_ allow a later project file to add one, we will support a new value for the package manager keys in the `"volta"` section: `false`. Setting `"yarn": false` will mean that no Yarn is available _and_ that a file later in the sequence with a `yarn` setting won't be able to override that removal of Yarn.
+
+Additionally, while we are processing this chain of project files, we will need to take care to detect accidental loops in the files pointed to by `"extends"` and exit early with a helpful error message (as opposed to entering an infinite loop of parsing and never finishing).
+
+Finally, in lieu of the current parse error for not setting `"node"` in the `"volta"` section, we should merge all of the project settings into a single project Platform, and then only show an error if none of the files specified a Node version.
+
+## Local Binary Resolution
+
+Currently, in order to run a project-local binary, we look for the executable in `<project root>/node_modules/.bin/`. With multiple project roots, this process will be almost the same: For each root in the sequence, we check for the binary in the expected place, and if it's found, then we stop and use that version. If we don't find it, we move on to the next root in the sequence.
+
+## Project-Local Hooks
+
+Similar to how we resolve the Platform, we will resolve project-local hooks by looking for `<project root>/.volta/hooks.json` for each project root. We will merge them together, with earlier entries having precedence over later ones, and then finally merge the result with the user-level hooks in the same way that we already do to merge project and user-level hooks.
+
+## Pinning Tools
+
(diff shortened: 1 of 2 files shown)


text/0000-monorepo-support.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Project-Local Hooks
+
+Similar to how we resolve the Platform, we will resolve project-local hooks by looking for `<project root>/.volta/hooks.json` for each project root. We will merge them together, with earlier entries having precedence over later ones, and then finally merge the result with the user-level hooks in the same way that we already do to merge project and user-level hooks.
+
+## Pinning Tools
+
+When a user runs `volta pin`, we will always make the changes to the _first_ `package.json` that we find. This means that if a user wants to `pin` a tool for the root project, they will need to first navigate to the root directory (out of any subprojects) and _then_ run `volta pin`. This gives users the most flexibility to control where they want to `pin` tools, without trying to guess which location they were intending or requiring interactivity.
```
rwjblue, line 57: Can we make the behavior of `volta pin` workspaces aware? Most of the time when you are working in a monorepo you are running your terminal in the top level project root. Assuming that is the case, then it seems like we could make `volta pin` notice that there is a `workspaces` property in that `package.json` and from that we can update all of the `package.json` files for you (with the right `volta: { extends: '../../package.json' }` bit.

Having to do it manually is fine (and the result with `extends` proposed here is much better), but IMHO we can still do **exactly** what the user wants 98% of the time...

text/0000-monorepo-support.md
```diff
@@ -0,0 +1,81 @@
...
+- The behavior of `volta pin` in subprojects vs the root project.
+
+Ideally, we should document all of these with specific examples to show how the different pieces interact. We could potentially even create an `examples` repo (or directory within the main repo) for users to reference.
+
+# Details
+[details]: #details
+
+## Multiple Roots
+
+The core change to Volta will be expanding our `Project` model to support a _sequence_ of project roots, instead of a single project root. We will support a sequence by adding the ability to put an `"extends"` entry inside the `"volta"` section of `package.json`. The value of that entry will be the path to another json file, which we will parse just like the original `package.json`. The directory that new file is in will be the next value in the sequence of project roots. Then, if that new file also has an `"extends"` entry, we will continue the same process, building up effectively a linked list of project roots to use for all of our project information.
```
rwjblue, line 33: This prose reads well, but I am struggling to fully understand what the actual contents of `volta` property in the `package.json`'s will be. Can you give a basic folder structure diagram (ala from the `tree` or `exa --tree` commands) and then specify what the contents would be for each `package.json`?

Something like:

---

For example, given the following folder structure:

```
.
├── package.json
└── packages
   ├── bar
   │  └── package.json
   └── foo
      └── package.json
```

The `package.json` files would have the following contents:

* `package.json`
```json
{
  "volta": {
    "node": "12.5.9",
    "yarn": "1.22.0"
  }
}
```

* `packages/bar/package.json`

```json
{
  "volta": {
    "extends": "../../package.json"
  }
}
```


* `packages/foo/package.json`

```json
{
  "volta": {
    "extends": "../../package.json"
  }
}
```

text/0000-monorepo-support.md
```diff
@@ -0,0 +1,81 @@
...
+
+Additionally, while we are processing this chain of project files, we will need to take care to detect accidental loops in the files pointed to by `"extends"` and exit early with a helpful error message (as opposed to entering an infinite loop of parsing and never finishing).
+
+Finally, in lieu of the current parse error for not setting `"node"` in the `"volta"` section, we should merge all of the project settings into a single project Platform, and then only show an error if none of the files specified a Node version.
+
+## Local Binary Resolution
+
+Currently, in order to run a project-local binary, we look for the executable in `<project root>/node_modules/.bin/`. With multiple project roots, this process will be almost the same: For each root in the sequence, we check for the binary in the expected place, and if it's found, then we stop and use that version. If we don't find it, we move on to the next root in the sequence.
```
rwjblue, line 49: I think this section would also be much easier to understand with a small diagram. I'm thinking something like:

---

Given the following folder structure:

```
.
├── package.json
└── packages
   ├── bar
   │  └── package.json
   └── foo
      └── package.json
```

When a previously `volta install`ed binary is executed from within the `packages/foo/` subdirectory we will look for a local installation of that binary in the following locations:

1. `packages/foo/node_modules/.bin/<binary name>
2. `node_modules/.bin/<binary name>
3. Fall back to our own global (sorry, I know that isn't the specific term we use, but I couldn't remember the right term)

charlespierce:
@rwjblue Thanks for the comments! On the `yarn: false` thing, I was thinking that it was related to having multiple roots, but you're right that we already do a merge and so it would be a new feature unrelated (though applicable to this use-case). I'll actually remove that entirely and make it a separate RFC.

Thanks for the suggestion on how to format some examples! I wanted to include some, but struggled a bit with how to format everything and eventually decided getting it out in some form was better, but I like that layout and should be able to make some!

For the `extends`, I'll update the text, but you're right that it's a path to another manifest file. It doesn't strictly _have_ to be a `package.json`, but it does need to be a JSON file with a top-level object that has a `volta` property. It also doesn't necessarily need to be a relative path (though it probably should be if you want to collaborate with other people).

text/0000-monorepo-support.md
charlespierce: Definitely, I like the tree layout to help make a compact example!

text/0000-monorepo-support.md
charlespierce: I agree we want to do what the user wants, but I'm also a little wary of a command that normally only affects a single `package.json` suddenly having knock-on effects in a bunch of different files. My concern is that if someone wants to pin the top-level _without_ updating the subprojects to have `extends`, then we don't really give a way to do that (apart from manual edits). Either that or we have to add a flag to the CLI that only applies if you are A) In the root of a Yarn Workspaces project and B) Trying to pin the top-level without making any changes to subprojects, which is a little to specific to be useful as a command-line flag, IMO.

That said, the benefits to getting started with a workspace (where we know we have a reverse mapping to the individual projects) are very nice, so it feels like it would be a good QoL improvement. I'll add it and we'll see if there are other comments / issues raised about how it works.

charlespierce:
Updated to improve the description in a few places and add examples to the `Details` subsections

dherman:
I'd like to add another option into the discussion, which is to have a less explicit linking model, closer to how I believe both Yarn and Cargo do workspaces. In this model, instead of `"extends"` being the way a nested package connects itself to the workspace, it would be the root manifest of the workspace that would indicate which nested packages are part of the workspace.

The reason I bring this up is that I'm concerned that with explicit backlinks, it requires the creator of each nested package to take responsibility for participating in the monorepo. For some monorepo use cases, there are separate teams responsible for the maintenance of individual packages vs the maintenance of the overall monorepo. With some sort of declarative language (typically, filesystem paths + globs) for include- and/or exclude-lists, a central team can manage the overall monorepo and establish conventions so that other teams can add new packages without having to know how to "opt in" to using the monorepo's configuration.

I mentioned this to @charlespierce yesterday and he brought up a few concerns to consider:
- Performance: This would likely require more filesystem crawling. Potentially expensive, but would be useful to do experiments to see if this is a problem in practice or not.
- Ability to use non-`package.json` manifests (as discussed in #282): I think this is solvable orthogonally, but worth thinking through.
- Nesting: Does this rule out the idea of a workspace within a workspace? Is that important? We weren't sure.

Nevertheless, I think it's worth discussing because we should think about how companies or large open source projects with large monorepos would be able to modularize the responsibility of managing the workspace configuration. My concern is that the RFC as-is diffuses responsibility for managing the workspace configuration to all sub-packages.

charlespierce:
@dherman Updated the critique to include your points about the modular responsibility. While I ultimately come to the same conclusion, given the nature of Volta, I think it's important to call out, thanks!

tujoworker:
Both Prettier and ESlint are supporting packages to extend with:

```json
  "prettier": "@scope/prettier-config-package",
  "eslintConfig": {
    "extends": "@scope/eslint-config-package"
  },
  "volta": {
    "extends": "../../volta-config-package/volta-config.json"
  }
```

For Volta, as for now, we have to have json file, either directly as a package.json or another json file.
Is that RFC about that?

I would love have this feature, because it makes handling of configuration in larger mono-repos, very much easier.

charlespierce:
Hi @tujoworker, this RFC right now is about the ability to use another JSON file to extend the Volta config. Can you expand a bit on the feature you're asking about? I'm not sure I see a clear path to Volta "extending" from packages, since Volta sits at a higher level than individual packages. One possible option would be to support a folder instead of a file, and then automatically append `package.json` to it, but I'm not sure that improves things much over the direct path.

The difficulty with using a package is that it creates a Catch-22: In order to install the dependencies, we need to know the Node / npm / Yarn version to run. But in order to know that, we would have to have dependencies installed so that we can "extend" from a package.

tujoworker:
Yeah, I will expand a bit; the thoughts are - to have the config only in a separate and dedicated package. 
Because workspace packages can be placed where ever they are defined in the workspace config, the paths to the json file can be different from package to package.
So, defining the Volta config inside an independent package - without any dependencies by itself - would make it both declarative and easy to refer to - in contrast to a path/folder.

But if this really is not possible because of the catch-22 scenario - then just ignore my thoughts / wishes.

rwjblue:
@charlespierce - This seems to be fully implemented, should we merge?

charlespierce:
@rwjblue Yeah, I think so. There are still some outstanding questions about the possibility of supporting some sort of automatic hierarchy, but I think that's more fine-tuning the existing implementation than completely rewriting it.

PR #45, opened 2020-07-16
Rethinking Global Packages

charlespierce:
> Update our handling of global package installs to use existing package manager idioms (e.g. `npm i -g`) with minimal mental overhead.

[Rendered RFC](https://github.com/charlespierce/rfcs/blob/global_packages/text/0000-global-packages.md)


text/0000-global-packages.md
```diff
@@ -0,0 +1,151 @@
...
+
+## Interception of Global Installs
+
+When a user runs `npm i -g <package>` or `yarn global add <package>`, instead of throwing a hard error as we currently do, we can make 2 small changes to the workflow but otherwise allow the command to run as expected:
+
+1. Use the `prefix` and `global-folder` options to redirect the install directory to the Volta-managed location. Package managers already provide support for changing the global install directory, so we can leverage that to make sure we always install into the appropriate Node major version cache. We should set these (if possible) using environment variables, so that they can still be overridden by command-line switches from the user in cases where they want to have control over the directories.
+
+2. After the install completes, take inventory of the packages that changed (whether installed, upgraded, or uninstalled) to update an internal manifest of all packages the user has globally installed.
+
+Additionally, if a user calls `volta install <package>`, we can perform the same steps and shell out to `npm i -g <package>` as if the user had typed that directly.
```
thoov, line 76: I wonder if we should just deprecate this during this RFC. We have already overloaded the `install` command npm global packages + "primitive" executables and this RFC seems like the appropriate time to deprecate this particular subcase.

text/0000-global-packages.md
```diff
@@ -0,0 +1,151 @@
...
+
+## Automatic Migration When Switching Node Versions
+
+In order to avoid requiring users to keep track of their global packages and manually re-install them whenever they switch Node versions, we can instead manage that for them. Using the manifest generated above whenever a package is updated, we will always know which tools the user has available to them. Then, whenever they switch to a new major version of Node (with `volta install node@<version>`), we can ensure that the full list of packages is made available by running the appropriate `install` commands.
+
+In order to be efficient at this change, we should keep a separate manifest for each major version of Node, so that we can perform a diff with the "primary" manifest and deteremine exactly which changes are needed to get the system in a working state.
+
+This process may take some time, so we should provide some visual feedback to the user to indicate what we're doing (and if possible, progress on how much is left to do). Additionally, we should have a command-line switch to allow users to skip the consolidation if they don't need it or want to make a quick change.
```
thoov, line 84: We should be very cautious around the dev UX of this. I could see an example where a user wants to upgrade their node version and all of a sudden the `volta install node` command takes a lot longer than expected (as it now could be now doing much more than just installing node). Food for though could be either to add some other command to upgrade the globals or to provide a Y/N prompt after the install if the user want to do the migration.

Additionally, we don't want any migration to ever cause an exit code > 1 if a singular migration fails. So we should make sure we make the migration as bullet proof as possible.

thoov: Another point here is it might be worth talking with some folks about the number of globals they have installed to get some kind of bench mark as to how long a typical migration might take.

We could also store the install times as apart of the manifest file and if the install times cross a certain threshold then we could require the user to agree before hand or alert them somehow what is happening in the background

text/0000-global-packages.md
charlespierce: I've gone back and forth a bit on this. On the one hand, reducing the overload of `volta install` is nice, but on the other, the ability to quickly spin up an environment with `volta install node yarn ember-cli` is _really_ nice.

Ultimately, I think that the discussion of changing the semantics of `volta install` is separate from this (though potentially dependent on this change), so it should be its own RFC. As long as `volta install` keeps the current behavior, I think it's useful to keep the package support around (even if it's mostly a pass-through to `npm i -g <package>`). When / if we decide to change around the commands, that's a better point to talk about deprecating support for packages directly through Volta.

text/0000-global-packages.md
charlespierce: The runtime of installing the separate packages is definitely a concern, and that's why I wanted to call it out here. Specifically, by "visual feedback" in this line, I'm thinking something like a new progress bar with a clear message of what Volta is doing. It definitely shouldn't be in the background or opaque, it should be very clear the user what we are doing at a high level.

I'm also wary of Y/N prompts for a couple of reasons: First, they break scripting without adding a `-y` flag of some king to make them non-interactive. Next, even for situations where the interaction itself is possible, they can really damage UX by being repetitive ("Yes, I _always_ want to update my packages, why do you keep asking?") or by happening after a delay when the user isn't looking at the screen. For that reason, I generally lean much more towards having a reasonable default and then a flag for unusual cases where that doesn't apply. To maintain Volta's features, it makes sense to have re-installing be the default behavior, while opting out is the special-case.

> Additionally, we don't want any migration to ever cause an exit code > 1 if a singular migration fails. So we should make sure we make the migration as bullet proof as possible.

Definitely agreed here! I intended to include something about that, I'll add it now. We should also warn the user if something fails and we expect a tool won't work.

charlespierce:
@thoov Good points all around. I definitely think when we have an implementation of this ready, we should do a beta / RC release and get some of our power users to test it out so we can get a feel for how much extra work is done during the migration process.

Updated some of the language and added a section around errors explicitly calling out that they shouldn't affect the success or failure of `volta install`.

text/0000-global-packages.md
```diff
@@ -0,0 +1,155 @@
...
+> npm i -g typescript@3.8.2
+```
+
+Finally, the user decides that working on Node 14 isn't working out, and they switch back to Node 12:
+
+```bash
+> volta install node@12
+```
+
+After ensuring Node 12 is available, Volta will once again check the manifest of global packages against the packages that are currently set up in the Node 12 bucket. In this case, Volta detects that we need to install `ember-cli@3.18.0` and downgrade `typescript` to `3.8.2`, so it runs `npm` commands internally to ensure those changes are made. When the process is complete, the user has the exact same tools that they did on Node 14.
```
haochuan, line 135: I'm curious in this case:
After switching back to `node@12`, let's just assume `ember-cli@3.18.0` is not compatible with `node@12`, and its min requirement is `node>=14`:

1. What will happen to the global `ember-cli` after switching back to `node@12`? For me personally I would expect some message to tell me this issue at least, like:

```
success: installed and set node@12.16.2 (with npm@6.14.4) as default
Checking global installed packages...
- [success] typescript@3.8.2
- [fail] ember-cli@3.18.0
    ember-cli@3.18.0 is not compatible with node@12.16.2, it requires `node>=14`. Please try downgrade `ember-cli` or upgrade `node` in order to use it
```

2. Most of the "good" packages will define the node version requirement in `package.json` via `engine`, and I assume we will also use it to check the compatibility. However, what if a package `foo` requires `node>=14`, but  don't have the `engine` defined in `package.json`(or wrong version)? After switching back to `node@12`, do we have better ways to tell if `foo` would work with `node@12`? And if user fails to run `foo` with `node@12`, do we have a way to let user know what's going on?

text/0000-global-packages.md
```diff
@@ -0,0 +1,155 @@
...
+* **Global packages should leverage Node module resolution** Our directory structure currently doesn't have any folders named `node_modules`, so the Node module resolution algorithm is completely unable to detect our packages, even from within a package's dependencies. There shouldn't be a need to completely reinvent the wheel here, we should do our best to leverage the existing resolution algorithm.
+
+## Overview
+
+The core of this approach is to take advantage of the existing behavior of package manager global installs like `npm i -g` and `yarn global add`. Those tools already provide the expected interoperability between global packages that we want and users have intuition around how they work. We can insert some customization before the global install runs to point it at a custom directory that Volta controls, so we can keep track of everything. Additionally, we can do record-keeping after the global install finishes to ensure that we keep track of which tools and binaries are installed.
```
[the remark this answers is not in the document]
charlespierce: It looks like it does, yes. Specifically, `pnpm` respects `NPM_CONFIG_PREFIX` in the same way that `npm` does, so setting that one environment variable should work for both. We may need to experiment with it slightly, to determine if the value should be exactly the same, but generally speaking, yes `pnpm` allows you to override the global install directory as well.

text/0000-global-packages.md
```diff
@@ -0,0 +1,155 @@
...
+
+## Sandbox Per Node Major Version
+
+The major issue preventing us from having a single persistent store of global packages is that Node major versions can represent breaking changes that cause a package to stop working. This is most notable with native modules, which are compiled against a specific Node Modules API version and need to be recompiled to work against a different Node. Since these types of changes are primarily limited to major version bumps in Node, we can keep a separate cache of global packages for each major version. This will be stored within the Volta directory, most likely under `tools/image/packages/<node version>`.
+
+Within a given Node Major bucket, we expect packages to continue to work without needing to be re-installed, so we can avoid needing to have a separate cache for every individual Node version. This also allows us to reduce the amount of time we spend re-installing packages on Node version switches.
```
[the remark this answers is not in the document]
charlespierce: That is the assumption, yes. I'll update the RFC to be a little more clear about the workflow, but the general idea would be we use `npm i -g` (or `yarn global add` if the user calls that) directly with the user's current default platform, adding the settings to redirect the install directory to one we control. So in your example, we would install `typescript` with Node 12.7.0, and then `ts-node` with Node 12.1.0, and assume that they work correctly together.

I don't know how common it is for packages to depend on specific features in minor versions of Node, though I suppose it's possible. If we want to work around that / ensure those issues don't come up, I suspect we'll need a different model entirely, as re-installing all of the packages for each Node individually is going to push the performance issues well beyond a reasonable use-case.

text/0000-global-packages.md
```diff
@@ -0,0 +1,155 @@
...
+
+## Handling Global Libraries
+
+Since libraries without binaries are already supported by `npm i -g`, this approach also includes that support without needing any modifications.
+
+## `volta list`
+
+Since this change represents a significant update of the internal model for packages, it will require the output of `volta list` around packages to be redesigned. Specifically, packages will no longer be associated with a specific platform any more, so we should be able to simplify the output to only show the package version (and whether it is the default version or a project-local version).
```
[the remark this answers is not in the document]
charlespierce: Good idea!

text/0000-global-packages.md
charlespierce: 1. I was thinking something like that, yes, and I really like that output so we can easily show the status of the installed tools.

2. Outside of `engines`—or more generally, the `npm i -g` command failing—I'm not sure of a good way to determine if a given package is compatible with a specific Node version or not.

charlespierce:
Updated to a new proposal that incrementally improves our existing model, instead of throwing it out wholesale.

text/0000-global-packages.md
```diff
@@ -0,0 +1,195 @@
...
+6. **Global packages should leverage Node module resolution** Our directory structure currently doesn't have any folders named `node_modules`, so the Node module resolution algorithm is completely unable to detect our packages, even from within a package's dependencies. There shouldn't be a need to completely reinvent the wheel here, we should do our best to leverage the existing resolution algorithm.
+
+## Overview
+
+The core of this approach is to incrementally satisfy the constraints to solve the outstanding issues. This will allow us to roll out improvements in steps, without having to make major, sweeping changes to the Volta implementation all at once. That flexibility will also allow us to iterate more quickly on the solution to make sure we cover the most common use-cases.
+
+## 0. Keep Existing Sandbox-per-Package Model
```
dherman, line 54: Curious why this is zero-indexed?

text/0000-global-packages.md
```diff
@@ -0,0 +1,195 @@
...
+
+## 2. Shared Global Package Directory
+
+In order to satisfy constraints 4 & 5, allowing global packages to `require` each other, we should create a single shared "globals" directory that we add to `NODE_PATH` when executing a global binary (and _only_ for global binaries, as normal Node commands wouldn't have access the globals anyway). That directory can have symlinks to all of the individual package directories, so that a `require` from any one of the packages will be able to access any of the others. Per the [Node documentation](https://nodejs.org/api/modules.html#modules_loading_from_the_global_folders), the `NODE_PATH` will only be searched if the regular filesystem process fails, so this should result in the correct precedence when resolving a global library.
```
dherman, line 85: Is this `"globals"` directory truly just one for all of Volta, or is it one per Node installation? If package `foo` is installed bound to Node 12 and package `bar` is installed bound to Node 14, calling into one another might not work at all (for example if `foo` has native components compiled to the Node 12 ABI).

text/0000-global-packages.md
charlespierce: Mostly because it wasn’t a change needed, just a call-out of how we already meet the one constraint.

text/0000-global-packages.md
```diff
@@ -0,0 +1,195 @@
...
+
+With the shared package directory available, we can then remove the restriction that we currently enforce that prevents installing global libraries.
+
+Finally, we should document the location of this shared package directory, so that if users want to include it in their own configurations (e.g. for one-off scripts), they can easily locate the correct directory.
```
dherman, line 89: If we expect people to use it in scripts it's probably a good idea to have a `volta` command (like `volta info globals-dir`) that can spit out the proper path, so that can be the "public API" for relying on the directory, making it easier for users and at the same time easier for us to potentially change the location in the future.

charlespierce: That’s a good call, it would definitely be good to expose that in the same way that `volta which` unwraps the shims.

dherman:
This seems very promising. I love that it preserves the benefit that once you install a global binary, it keeps working even if you change your Node version. The part I'm still fuzzy on is the precise difference between (a) sandboxing every global package but making them visible to each other for `require` vs (b) not sandboxing global packages and just installing them directly in a Node version.

text/0000-global-packages.md
charlespierce: The thought was to have a single directory, which potentially could cause problems of packages across Node versions not working.

The flip side of that, however, is if we isolate by Node version then there’s the potential for confusion when `require` fails to find a package that _was_ installed.

My sense is that 1) native modules are relatively rare and 2) Default versions change infrequently enough to minimize the potential breakages.

Another concern is even if we isolate packages by Node major version, there’s no guarantee it’ll work across versions (different features / bug fixes within a given major version). So there is always going to be the potential for breakage that we’ll need to deal with.

text/0000-global-packages.md
dherman: PResentationally, maybe you could make that a little more obvious, like "Step 0 (done!) Keep existing Sandbox-per-Package Model or something like that. NBD tho!

text/0000-global-packages.md
dherman: It seems worth calling out somewhere that we expect native modules to be rare in global packages, and that this means that if people _do_ happen to have Node-version-sensitive global modules, they'll need to be careful about what version of Node they're using. That seems reasonable to me, it's just worth being explicit about it in the RFC.

Also, note when it comes to our (separate) work on `npm link` etc, we may not be able to make the same assumption. In local packages, native modules are more common. Not relevant for this RFC, I just want to make sure we don't over-learn from the assumptions we make here and apply them elsewhere where they may not hold.

Your reasoning here sounds plausible to me but should it be made more explicitly in the RFC?

text/0000-global-packages.md
charlespierce: Added a "(Complete!)" to this line, good thinking!

text/0000-global-packages.md
charlespierce: Added a "Critique" section covering that and noting the need to clearly document the potential for incompatibility and how to fix it. Might be worth creating a "Troubleshooting" or something section in the docs to cover FAQ type questions and known issues.

charlespierce: Also definitely agree that `npm link` definitely needs to be more explicit about the Node versions. Luckily in that case (at least if both projects are using Volta), we can detect a mismatch and show a warning at the time the `link` is created.

text/0000-global-packages.md
charlespierce: Included a call-out that a new command would be helpful for this.

text/0000-global-packages.md
mikrostew: We may be able to detect when a user is installing a package that contains native modules, and give some kind of appropriate warning: https://stackoverflow.com/questions/36032250/how-to-find-native-modules-in-a-node-js-dependency-tree

charlespierce: Oh great find @mikrostew! I’ll add a note that we should show a note when installing a native module to remind the user that they should be aware of Node version compatibility

charlespierce:
@dherman Updated to clarify the sections you had questions about, thanks again for the feedback!

PR #48, opened 2022-02-17
Supporting Yarn Berry

charlespierce:
> Roadmap of the necessary changes to support Yarn Berry (Yarn 2.0+)

[Rendered](https://github.com/charlespierce/rfcs/blob/yarn_berry/text/0000-yarn-berry-support.md)

### DIFF
--- a/text/0000-yarn-berry-support.md
+++ b/text/0000-yarn-berry-support.md
@@ -1,0 +1,65 @@
+- Feature Name: yarn_berry_support
+- Start Date: 2022-02-16
+- RFC PR: (leave this empty)
+- Volta Issue: https://github.com/volta-cli/volta/issues/651
+
+# Summary
+[summary]: #summary
+
+Roadmap of the necessary changes to support Yarn Berry (Yarn 2.0+)
+
+# Motivation
+[motivation]: #motivation
+
+We currently only support legacy Yarn 1.x installation and pinning. While Yarn 2.x had a lot of infrastructure churn that made it difficult to support, Yarn 3.0+ is more stable and provides all of the pieces that we need to add support. To maintain compatibility as Yarn continues to evolve, we should support Yarn Berry.
+
+# Pedagogy
+[pedagogy]: #pedagogy
+
+The ideal outcome with this change is that there will be very little we need to teach users. Newer versions of Yarn will be supported and users will be able to continue to use Volta the way they always have, without needing to worry about specific details of their setup.
+
+The main detail we will need to communicate is the lack of support for the 2.x major version of Yarn (only supporting Yarn 1.x and Yarn 3.0+).
+
+# Details
+[details]: #details
+
+## Skip Yarn 2.x Support
+
+As mentioned above, the initial release of Yarn Berry involved a lot of changes that were difficult to track and didn't fit with Volta's model for handling package installation. The biggest blocker is that the distribution package on npm wasn't created until the last version of Yarn 2. Before that, there wasn't any reliable way to fetch the Yarn binary and make it available within Volta.
+
+Since the only viable distribution is the _final_ version of Yarn 2 and Yarn has since released Yarn 3, it makes the most sense to skip support for Yarn 2 alltogether and jump straight from Yarn 1 to Yarn 3. That will allow us to detect the supported versions very easily, using only the major version number rather than a bespoke list of "allowed" versions. We should provide a useful error message if a user attempts to install or pin Yarn 2, directing them instead to use Yarn 3 for the most up-to-date experience.
+
+## Installation
+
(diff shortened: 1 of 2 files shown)


text/0000-yarn-berry-support.md
```diff
@@ -0,0 +1,65 @@
...
+
+### Hooks
+
+One outstanding question is how we want to handle the hooks for installs. Since the Hooks format was created before we switched Yarn to use the npm registry, the logic is rather complicated:
+
+* If we aren't using hooks, then we fetch from npm and expect the data to match the npm package format
+* If we _are_ using hooks, then we fetch from the location specified by the hook and we expect the data to be in the old GitHub API format
+
+With the addition of Yarn 3 pulling from a different package (though still using the npm registry), this could get more complicated. My current suggestion is that we keep the hooks format the same even for Yarn Berry releases. This will mean that teams using hooks will have to make sure the correct package is available at the appropriate URLs themselves. Then, in a future change where we update the hooks format, we can migrate all of Yarn to use the npm format and normalize the behavior.
```
mikrostew, line 43: A small change to help with this could be an optional `"format"` field in the `"yarn"` section of hooks.json, which would take `"npm" | "github"`, like:

```
    "yarn": {
        "index": {
            "template": "http://example.com/yarn/{{os}}/{{arch}}/yarn-{{version}}.tgz"
        },
        "latest": {
            "prefix": "http://example.com/yarnpkg/"
        },
        "distro": {
            "bin": "~/yarn-distro"
        },
        "format": "npm"
    }
```

That would specify the format Volta should expect. Leaving this out would default to the current behavior of using the Github API format.

text/0000-yarn-berry-support.md
```diff
@@ -0,0 +1,65 @@
...
+[motivation]: #motivation
+
+We currently only support legacy Yarn 1.x installation and pinning. While Yarn 2.x had a lot of infrastructure churn that made it difficult to support, Yarn 3.0+ is more stable and provides all of the pieces that we need to add support. To maintain compatibility as Yarn continues to evolve, we should support Yarn Berry.
+
+# Pedagogy
+[pedagogy]: #pedagogy
+
+The ideal outcome with this change is that there will be very little we need to teach users. Newer versions of Yarn will be supported and users will be able to continue to use Volta the way they always have, without needing to worry about specific details of their setup.
+
+The main detail we will need to communicate is the lack of support for the 2.x major version of Yarn (only supporting Yarn 1.x and Yarn 3.0+).
```
mikrostew, line 21: We will also need to communicate how users should handle an upgrade from Yarn 1.x --> Yarn 3.x with Volta.

The [yarn migration docs](https://yarnpkg.com/getting-started/migration#step-by-step) are a good resource for this, but Volta users should not need to run `npm install -g yarn` or `yarn set version berry` (although we could potentially intercept those commands and Do The Right Thing ™️ )

text/0000-yarn-berry-support.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Hooks Format
+
+The current suggestion for the hooks is the easiest from an implementation standpoint, however it offloads that work onto users that are taking advantage of hooks. However, our Hooks behavior is already not ideal and adding more complexity to it feels like it would make things even more difficult to understand. Separate from this issue, we should investigate an improved Hooks V2, with a new file format and better ergonomics, taking into account everything we learned from the original hooks implementation. At that time, we can define a more appropriate API for the Yarn hooks. That said, I don't believe our support for Yarn Berry should be blocked on redesigning Hooks, which is likely a much bigger issue overall.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- Should we consider Corepack support as part of support for Yarn Berry? ([Corepack](https://nodejs.org/api/corepack.html) is Yarn's currently-recommended way of installing / managing Yarn versions)
```
mikrostew, line 65: Would this be doable as a followup to support for Yarn Berry?

mikrostew (commented):
Very excited to see this starting to happen!

text/0000-yarn-berry-support.md
charlespierce: Great suggestion! That would make the different formats more clear / explicit in a backwards-compatible way. It would also enable teams to use the `npm` format for all versions of Yarn, even the older 1.x ones.

text/0000-yarn-berry-support.md
charlespierce: Great call-out, we can definitely document how to use newer Yarn versions with Volta. AFAICT, Yarn Berry doesn't have "global" commands, so really it should be `volta pin yarn@3` and that's all you need.

text/0000-yarn-berry-support.md
charlespierce: Yeah, this can definitely be separate, especially if we have documentation on the "Volta way" of making it available.

chriskrycho: Yeah, I think we need to tackle Corepack separately, as it's a very different bundle of concerns and is sort of notionally a competitor to *part* of what we do in Volta.

chriskrycho (approved):
👍🏼 Let's do it.

charlespierce:
We already have a good chunk of this implemented (thanks @mikrostew!), moving this final comment period before merging soon since we're very likely to release this generally unless there's a major issue discovered.

nfantone:
@charlespierce Apologies for the ping, but is there anything blocking you from releasing this? Can I help in any way?

Relequestual:
Hey @charlespierce, how long is the "final commit period"? We are nearly a year on since you tagged this PR =]

chriskrycho:
This probably just needs to actually be merged! 😁 We shipped Yarn 3 support quite some time ago!

PR #49, opened 2022-03-29
Delete LICENSE

chantik3552:


chriskrycho:
Hi @chantik3552 – not sure what your intent here was, but we'll be keeping the LICENSE. Thanks.

PR #52, opened 2024-02-22
Updating `volta pin` to support a `--default` flag

charlespierce:
Proposal for updating `volta pin` to add a `--default` flag that replaces the need for `volta install`, combining all of the pin operations into a single `volta pin` command.

[Rendered](https://github.com/charlespierce/rfcs/blob/volta_use/text/0000-volta-use.md)

### DIFF
--- a/text/0000-volta-use.md
+++ b/text/0000-volta-use.md
@@ -7,7 +7,7 @@
 
 [summary]: #summary
 
-Adding a new command, `volta use`, which combines the effects of `volta install` and `volta pin` into a single command that is more intuitive and better matches the mental model of Volta.
+Updating `volta pin` by adding a `--default` flag to support pinning the _default_ version of a tool (instead of the local project version). Combining the behavior of `volta pin` and `volta install` into a single `pin` command that is more intuitive and better matches the mental model of Volta.
 
 # Motivation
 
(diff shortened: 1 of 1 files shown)


text/0000-volta-use.md
```diff
@@ -0,0 +1,103 @@
...
+
+## `volta default` or `volta global`
+
+In the vein of replacing the verb for `install`, I also considered `default` or `global` for the command. The major issue with those is that they're adjectives, rather than verbs, so it's not immediately clear what the command means. Does `volta default node` mean "choose a default node" or "list the current default node"? That confusion is ultimately why I decided against using either of these for the command.
+
+## `volta set`
+
+Another possibility for the command would be to make it more verbose, e.g. `volta set default node`. That solves the problems around `volta default` or `volta global` by adding a verb to make it very clear what you are doing. However, it makes the commands a fair bit more cumbersome to use, which I feel is ultimately a net negative compared to a single command.
```
charlespierce, line 97: One idea here may be to use `volta set` as the new command, rather than `volta use`. That makes it even more clear that the important part of the operation is setting the version, however it breaks from users expectations from other tools.

text/0000-volta-use.md
```diff
@@ -0,0 +1,103 @@
...
+[details]: #details
+
+## Command definition
+
+```
+USAGE:
+    volta use [FLAGS] <tool[@version]>...
+
+FLAGS:
+    -g, --global    Select the global version of a tool, which is used everywhere outside of pinned projects
```
chriskrycho, line 43: Historically, at least, we have always used the language of “defaults” rather than “globals” to distinguish that Volta’s mental model is a bit different.

```suggestion
    -d, --default    Select the default version of a tool, which is used everywhere outside of pinned projects
```

text/0000-volta-use.md
```diff
@@ -0,0 +1,103 @@
...
+    -p, --pin       Pin a version of a tool to the local project, creating a Volta entry in package.json
+
+ARGS:
+    <tool[@version]>... Tools to make available, like `node@lts` or `yarn@^1.14`
+```
+
+## Behavior
+
+The behavior of `volta use` will ultimately be the same as the current `volta install` or `volta pin`. However, when run without flags, it will be contextual, based on where the command is being run. So if a user runs `volta use node@18` in a pinned project, the project will be updated to reflect the newly selected version. If they instead run `volta use node@18` _outside_ of a pinned project, then the default Node version will be updated. This ensures that running `volta use node@version` will always update the version that is currently active, meaning the next run of `node` will use that updated version.
```
chriskrycho, line 52: I am mulling on the design here in general and my reaction to it, and I could probably go either way on the overall collapsing of `install` and `pin` into a single `use` verb, but I *deeply* dislike this kind of contextual behavior as a user of this kind of tool. The reason I dislike it is that I want the same command to behave the same way given the same arguments. This feels like a case of “global mutable state”, i.e. the directory I am running in and whether it happens to have had a `package.json` file added to it or not, affecting the behavior of a function call, and I personally prefer to *avoid* my `static mut` values. 😆 It seems particularly iffy to me in that it *also* gets changed by whether there is a `package.json` in a directory anywhere between me and my home directory at the time it happens.

Consider a hypothetical folder layout:

```
/users/chris/code/
  DefinitelyTyped/
    DefinitelyTyped/
      package.json
      types/
        react/
          package.json
          ...
  volta-cli/
    package.json # notice this file
    rfcs/
    volta/
      src/
      some-node-support-script/
      ...
```

I will get completely different behaviors from `volta use node@lts` in each of these directories:

- `/users/chris/code`: sets the default
- `/users/chris/code/DefinitelyTyped`: sets the default
- `/users/chris/code/DefinitelyTyped/DefinitelyTyped`: sets the DT project
- `/users/chris/code/DefinitelyTyped/DefinitelyTyped/types`: sets the DT project
- `/users/chris/code/DefinitelyTyped/DefinitelyTyped/types/react`: sets the React project w/in DT
- `/users/chris/code/volta-cli/volta/`: sets the `volta-cli` default
- `/users/chris/code/volta-cli/volta/some-node-support-script/`: sets the `volta-cli` default

This is all internall consistent, and it is comprehensible, but I don’t find it *obvious*, specifically because it is *so* non-local. Scripts could of course make sure to always do the right thing by explicitly passing `--pin` or `--global`, but I do not like that they *need* to.

Also, I note don’t think we have any other command that works quite that contextually. `volta list` is close, but it just shows you the full hierarchy or your tools. If you are not in a project, you *only* get default/global-installed tools listed, but you will get those listed, with our shadowing rules applied, when you *are* in a package as well.

🤔

Thinking out loud a bit here: What if we more or less “renamed” `pin` to `use` and added `--default` to it? That eliminates all the ambiguity above, always behaves the same way, and still gets us all the *conceptual* simplification of having a single command that does what it says on the tin. It comes with the slight extra typing cost of `volta use --default node@lts` instead of “just working”. However, I think that tradeoff is probably worth it.

I would not fight *especially* hard about this, and definitely would not block on it, but I definitely prefer this design. Thoughts?

text/0000-volta-use.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Changing only the verb for `install`
+
+One alternative would be to only change the verb for `install`, leaving `pin` completely untouched. This would be the least disruptive change, however it's very dependent on the verb / command that we choose. With the proposal of `volta use`, it would feel very unintuitive to run `volta use node@21` in a project, and then immediately get a message that it's set as the global but your local settings override. As a user, I would expect `volta use node@21` to mean the next call to `node --version` to display Node 21 is running. In order to make that work as users expect, we need to unify the two so that `volta use` _does_ actually change the Node for the current context.
+
+Another difficulty of the unified behavior is that it's more difficult to explain when users run into edge cases. Right now there's a clear separation between "local" and "global" with two different commands. Unifying them means we'll need to clearly document how `volta use` behaves in each situation, and how to override that (with the `--global` or `--pin` flags).
```
chriskrycho, line 89: This is basically exactly my point above. 😂 I take it to be the most significant drawback, and I think it is the thing which really mashes my buttons.

text/0000-volta-use.md
```diff
@@ -0,0 +1,103 @@
...
+
+## `volta default` or `volta global`
+
+In the vein of replacing the verb for `install`, I also considered `default` or `global` for the command. The major issue with those is that they're adjectives, rather than verbs, so it's not immediately clear what the command means. Does `volta default node` mean "choose a default node" or "list the current default node"? That confusion is ultimately why I decided against using either of these for the command.
+
+## `volta set`
+
+Another possibility for the command would be to make it more verbose, e.g. `volta set default node`. That solves the problems around `volta default` or `volta global` by adding a verb to make it very clear what you are doing. However, it makes the commands a fair bit more cumbersome to use, which I feel is ultimately a net negative compared to a single command.
```
chriskrycho, line 97: 👍🏼 to these; I agree quite strongly!

text/0000-volta-use.md
```diff
@@ -0,0 +1,103 @@
...
+
+# Unresolved questions
+
+[unresolved]: #unresolved-questions
+
+- How do we handle the fact that you can install a 3rd-party package globally, but for a pinned project, we require users to install them normally using their package manager of choice? That is, what should we do when a user runs `volta use typescript` while they are in a package that is pinned?
```
chriskrycho, line 103: I see two options here:

1. Keep `install` around for *only* this, deprecating its use for installing runtimes.
2. Get rid of `volta install` entirely, and just teach people to use the package manager’s global installation tool (but have docs explaining it).

I personally understand why we added support for package-manager-based global installations but also think it significantly muddied our mental model *and* complicated things for end users, since it *doesn’t* work like the regular command does. Given that, it is probably not surprising that I like (1) here better. However, as things stand, I think (2) might actually be the cleaner path, my own personal preferences be darned. 😂

text/0000-volta-use.md
```diff
@@ -0,0 +1,103 @@
...
+  - If we are in a pinned project (one that already has a `"volta"` key in `package.json`) -> use `volta pin` behavior.
+  - Otherwise, use `volta install` behavior.
+
+# Critique
+
+[critique]: #critique
+
+## Confusion around omitting version
+
+One nice part about the existing `volta install` and `volta pin` commands is that they naturally work well with choosing a tool without selecting a version. For example, running `volta install node` makes intuitive sense as "pick whatever version and make it available." By contrast, `volta use node` doesn't immediately (to me, at least!) jump out as _picking_ a version of Node. `volta use node@lts` or similar, _do_ make sense, but omitting the version feels a little more awkward with `use` as the verb.
```
chriskrycho, line 83: One alternative this suggests is to actually *not* change `pin` to `use`, and just to stick with `pin`, but add the `--default` flag to it. That eliminates this oddity (which is, you are right, fairly significant!), and also gets rid of the need to introduce a new verb.

One thing that makes me think of: given that part of the motivation is that `pin` does not *necessarily* imply fetching, and that we *do* have a `fetch` command, what if we added a `--no-fetch` flag to `pin` as well? That feels to me like a really clear mental model:

> `pin` sets the version you ask for, and as a convenience, also fetches it if you don’t have it. If you want to `pin` a version *without* fetching it, you can run `pin --no-fetch`.

One nice thing that could unlock is the ability to set a version in a nice way while working offline. That would make it easy to use `volta pin` in e.g. project boilerplate generation without a network connection, for example. I have no idea if anyone wants that latter capability, but having it work offline, e.g. on a plane, even if that means I won’t be able to run using that, seems nice in some ways.

text/0000-volta-use.md
charlespierce: To start off, I agree that I don't like tools doing surprising or unintuitive things for inscrutable reasons. Balancing against that, however, are a couple points:

1. I _really_ dislike when tools give an error message that makes it clear they know what you intended, instead of doing the thing. We already have that a little when running `volta pin` outside of a package, but unifying the verbs but then having errors on `volta use npm@latest` that say `Did you mean "volta use --default npm@latest"` feels like going in the wrong direction.
2. We do have contextual behavior in the app already: The shims themselves! The core behavior of Volta is to provide this kind of contextual execution based on where you are when running a command.

That said, the point about `volta use node` is a great callout, since the "pin" case of that is the one entry point to _writing_ the `"volta"` key into an otherwise un-volta-fied `package.json` in the first place. The possibility of writing a config to a file _somewhere_, but it not being immediately obvious where, definitely feels like it will lead to hard-to-debug issues.

Also thinking out loud: I wonder if it would be clearer if we migrated to using a `.voltarc` for our "in-package" state and then made `volta use` only _create_ the file with a flag? That way the impact of calling `volta use` would be limited to modifying the existing `.voltarc` files, which very clearly indicate a connection with Volta.

text/0000-volta-use.md
charlespierce: Yeah, I think I would lean towards (2) given that we have the global interception and the two are generally the same (`volta install {package}` ultimately uses `npm i -g` internally anyway). That would also let us fully deprecate `volta install` while also simplifying `volta use` to be exclusively for defining the toolchain.

text/0000-volta-use.md
charlespierce: I still need to let it rattle around my brain a bit, but the more I think about it, the more I'm coming around to the idea of `volta pin --default`. I do really like the verb making it clear what's happening, and adding a `--no-fetch` option would further cement the mental model.

text/0000-volta-use.md
chriskrycho: Yeah, I also share (1) though I also have come around on that over time because it's really easy for a tool to do the *wrong* thing on the assumption that the interpretation there is the correct one, and saying "You *might* have meant this; if so here's how to do it" increasingly seems better to me over time.

On (2) I thought I wrote as much above, but I didn't! I must have thought of it *later*. Shower mode brain or something. 😂 I think the difference, to me, is in the weird and admittedly extremely fuzzy distinction between “I am proactively trying to do something with Volta” in the `volta use` (or `pin` or whatever) case vs. “I am passively letting Volta act transparently on my behalf while doing something *else*” in the `node some-script.js` case. That is, again, *exceedingly* fuzzy, but I think it is part of what is motivating my intuition here. Take it for what it is worth. 😝

text/0000-volta-use.md
charlespierce: It's slow-moving, but letting this ruminate on some of my recent runs, I really like keeping `pin` as the verb. It already clearly indicates the important part of the operation: pinning the version to use. When I get some free time during the long weekend, I'm planning to update this RFC to reflect that approach.

The main sticking point I have is that I think we want a good "short" flag for pinning the default version to make it short to type out, especially for the "Getting Started" guide. Using `--default` as the long flag doesn't work great, because the short version `-d` intuitively makes me think of a "delete" operation, rather than a "default".

While I know we've generally used `default` instead of `global`, I wonder if using `--global` here would make sense. It would align with people's experiences with package managers to make it clear what `volta pin -g node` means. Also, part of Volta's value prop is that it makes "globals" like Node, npm, and 3rd-party packages safe to use by delegating to the local version when it's available.

charlespierce:
Suggestion from the [main Volta repo](https://github.com/volta-cli/volta/issues/1787): Having `volta use` be an alias for this combined command would be useful for people switching from / used to `nvm`, since `nvm use` is used to change Node versions. It's not _exactly_ the same semantics, so we might want to consider how different it is and how difficult it will be to transition, but I think if we combine the commands it would be a good time to look at making `use` another alias for the same command.

chriskrycho:
I definitely don’t have an objection to `use`, @charlespierce; do you remember why we didn’t choose that in the first place? (I believe that decision preceded my working on the project, but it has been a *minute*, so one way or another I simply do not recall.)

text/0000-volta-use.md
charlespierce: Ultimately I agree that the contextual behavior has too many corner cases & opportunities for really poor UX. Updated the proposal to always use `volta pin` but with an explicit flag to indicate pinning the default version.

text/0000-volta-use.md
charlespierce: Finally got around to updating this RFC to the proposal of `--default`. I went with `-d` and `--default`, but also left the `--global` option as one of the critiques. I'm fairly ambivalent, and maybe there's a secret third thing we could find that would work as a better adjective for the behavior.

charlespierce:
Updated to a new proposal of continuing to use `volta pin` with an explicit flag (currently `--default`) for pinning the default version.

rwjblue:
@charlespierce - I really like this last round of updates, definitely makes more sense to me this way.

charlespierce:
> I definitely don’t have an objection to `use`, @charlespierce; do you remember why we didn’t choose that in the first place? (I believe that decision preceded my working on the project, but it has been a _minute_, so one way or another I simply do not recall.)

I don't have a deep recollection, but I think we actually _did_ have `notion use` as a command very early on (that's how long ago it was, prior to the rename). I can't find the git history of why we removed it, likely because it caused more confusion due to the differences with `nvm`

PR #54, opened 2025-03-04
RFC for support of .node-version file

andreassjoberg:


text/0000-dot-node-version-file-support.md
```diff
@@ -0,0 +1,63 @@
...
+
+# Summary
+[summary]: #summary
+
+Introduce support for a `.node-version` file, enabling it as a fallback source for determining project-specific Node versions. If node version is not specified in `package.json`, fallback to `.node-version` before using global defaults.
+
+# Motivation
+[motivation]: #motivation
+
+The aim is to align Volta with other popular Node version managers like fnm, which support `.node-version`. This enhances flexibility and user experience by allowing developers to specify their preferred Node versions in a widely recognized format.
```
rwjblue: 2/5 (minor preference, non-blocking) 

Maybe add a few more references (so it's clear it's not just `fnm`), also cross link to the tools websites/repos (so we can poke at them as we review).

text/0000-dot-node-version-file-support.md
```diff
@@ -0,0 +1,63 @@
...
+- Conflicting versions between `.node-version` and `package.json`.
+- Interaction with global Node version settings.
+
+# Critique
+[critique]: #critique
+
+By supporting .node-version, Volta will provide users with more flexibility in how they manage their Node environments, while maintaining its commitment to backward compatibility. This RFC invites feedback on the proposed implementation and any potential edge cases that might arise.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
```
rwjblue, line 99: A few open questions: 

- Should we also attempt to support `.nvmrc` (nvm doesn't support [`.node-version`](https://github.com/nvm-sh/nvm/issues/794))? (I think the answer is no, but I just think we shouldn't ignore the question)

text/0000-dot-node-version-file-support.md
```diff
@@ -0,0 +1,63 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+The aim is to align Volta with other popular Node version managers like fnm, which support `.node-version`. This enhances flexibility and user experience by allowing developers to specify their preferred Node versions in a widely recognized format.
+
+# Pedagogy
+[pedagogy]: #pedagogy
+
+Existing node developers are likely already familiar with other versioning methods like `.nvmrc`. Introducing support for `.node-version` aligns Volta with other node version managers.
```
rwjblue: 2/5 (minor preference, non-blocking) 

If we aren't planning on adding support for `.nvmrc`, we probably shouldn't mention it here?

text/0000-dot-node-version-file-support.md
```diff
@@ -0,0 +1,63 @@
...
+
+# Details
+[details]: #details
+
+## Inheritance from current solution
+
+The new feature will follow the same logic currently applied when reading a node version from `package.json`. This ensures consistency across different configuration files.
```
rwjblue: 3/5 (strong opinion: non-blocking)

> The new feature will follow the same logic currently applied when reading a node version from `package.json`.

I'm not 100% sure what this is referring to. It's probably best to define the current rules, instead of saying "follows the same logic currently applied".

text/0000-dot-node-version-file-support.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Standardized Approach
+
+This implementation should mirror the behavior of other Node version managers, ensuring Volta remains competitive and user-friendly.
+
+## Fallback Hierarchy
+
+Prioritize `.node-version` as fallback for determining Node versions.
+If volta is not defined in `package.json`, check `.node-version` before proceeding with global defaults.
+Continue with existing behavior if neither file specifies a version.
```
rwjblue: 4/5 (serious issue: blocking)

As written, this _feels_ backwards. It would be better to actually define the lookup precedence directly (instead of talking in terms of fallbacks).

Something like (but with better/more words):

```
The Node version lookup process will be:

1. Resolve the `package.json` and check it's `volta.node` field
  - if present, use that explicit node version
2. Check for the presence of `.node-version`
  - if present, use the value specified
3. Fallback to the default active `node` toolchain
```

text/0000-dot-node-version-file-support.md
```diff
@@ -0,0 +1,63 @@
...
+
+The aim is to align Volta with other popular Node version managers like fnm, which support `.node-version`. This enhances flexibility and user experience by allowing developers to specify their preferred Node versions in a widely recognized format.
+
+# Pedagogy
+[pedagogy]: #pedagogy
+
+Existing node developers are likely already familiar with other versioning methods like `.nvmrc`. Introducing support for `.node-version` aligns Volta with other node version managers.
+
+# Details
+[details]: #details
```
rwjblue, line 23: 5/5 (fundamental issue: blocking)

There are a few things missing from details section, that I think are required:


1. AFAICT, there is no spec so we *must* define what the contents of the `.node-version` file actually are. 
2. What happens if we find a malformed `.node-version` (e.g. one that specifies a range instead of an explicit version)

rwjblue (changes requested):

andreassjoberg:
@rwjblue Thank you for very good comments!

I have updated the RFC. Please have another look and let me know if anything needs clarification.

text/0000-dot-node-version-file-support.md
```diff
@@ -6,34 +6,56 @@
...
+```sh
+$ node --version
+v20.18.2
+$ node --version > .node-version
+```
+
+It is recommended to support optional leading `v` and any line ending [[reference](https://github.com/shadowspawn/node-version-usage#suggested-compatible-format)].  
+Allowing a leading `v` is common and gives a nice symmetry with `node --version`.  
+Allowing any line ending makes it easier for users and especially Windows users to create a compatible file.
+
```
rwjblue, line 46: These are great points! Let's just incorporate them above and have the text be more declarative (as in "we will support X"). Something like:

```
### Compatible File Format

The contents of `.node-version` will be:

- optionally a leading `v`
- three part numeric version (e.g. 20.18.2)
- optionally a trailing newline (either unix style or windows style line endings are allowed)

Any content found in addition to the above, will result in a an error message explaining that the `.node-version` file is malformed.
```

rwjblue (commented):
These are great updates, thank you!

text/0000-dot-node-version-file-support.md
```diff
@@ -0,0 +1,87 @@
...
+The aim is to align Volta with other popular Node version managers like [fnm](https://github.com/Schniz/fnm), [n](https://github.com/tj/n) and [asdf](https://github.com/asdf-vm/asdf-nodejs), which support `.node-version`. This enhances flexibility and user experience by allowing developers to specify their preferred Node versions in a widely recognized format.  
+A list of supporting products can be found here: <https://github.com/shadowspawn/node-version-usage>
+
+# Pedagogy
+[pedagogy]: #pedagogy
+
+Existing Node developers are likely already familiar with different Node versioning methods. Introducing support for `.node-version` aligns Volta with other Node version managers.
+
+# Details
+[details]: #details
```
rwjblue, line 23: 3/5 (strong opinion: non-blocking)

Would you mind adding a section to explain the error message we want folks to get when they have a malformed `.node-version` file? 

I'm thinking something along the lines of this:

```markdown
### Invalid `.node-version`

When an invalid `.node-version` is found (see < cross link compatible file format section? >), Volta will emit an error indicating (as best as we can tell) what is wrong with the file along with it's file path (implementation note: this could either be relative path or absolute path).

Primary error cases:

- Invalid semver range (e.g. `lts/*` or `20`)
> An invalid semver range (`lts/*`) was detected in `./.node-version`. Please specify an exact version number (e.g. `1.0.0`).
- Unsupported content (leading content, trailing content, etc)
> Unsupported content (at line 2, column 1) was detected in `./.node-version`. The `.node-version` should contain only a valid semver version number.
```

Feel free to tweak and make sound better as you see fit though!

andreassjoberg:
Thanks again @rwjblue! I've updated the file once again. 👍

text/0000-dot-node-version-file-support.md
```diff
@@ -0,0 +1,101 @@
...
+
+# Summary
+[summary]: #summary
+
+Introduce support for a `.node-version` file, enabling it as a fallback source for determining project-specific Node versions. If Node version is not specified in `package.json`, fallback to `.node-version` before using global defaults.
+
+# Motivation
+[motivation]: #motivation
+
+The aim is to align Volta with other popular Node version managers like [fnm](https://github.com/Schniz/fnm), [n](https://github.com/tj/n) and [asdf](https://github.com/asdf-vm/asdf-nodejs), which support `.node-version`. This enhances flexibility and user experience by allowing developers to specify their preferred Node versions in a widely recognized format.
```
rwjblue:
```suggestion
The aim is to align Volta with other popular Node version managers like [mise](https://github.com/jdx/mise), [fnm](https://github.com/Schniz/fnm), [n](https://github.com/tj/n), and [asdf](https://github.com/asdf-vm/asdf-nodejs) which support `.node-version`. This enhances flexibility and user experience by allowing developers to specify their preferred Node versions in a widely recognized format.  
```

rwjblue (approved):
This is looking really good! Thank you for working on it @andreassjoberg!

text/0000-dot-node-version-file-support.md
andreassjoberg: Fixed this!

text/0000-dot-node-version-file-support.md
mb21: Is this unresolved question blocking merging this PR? If so, I would propose to remove it, and add a sentence somewhere saying that we currently don't support `.nvmrc` file, and that support could be later added if deemed a good idea.

yvbeek:
This looks great, could we have this merged in?

